PR #49, opened 2024-10-18
Add marine infrastructure dataset and model config

favyen2:
This adds dataset/model configs for solar farm and wind turbine models, along with corresponding inference pipelines.

It also:
* Moves convert_satlas_webmercator_to_rslearn to one_off_projects since it's no longer needed now that the conversion of the datasets (from old format + WebMercator projection to new rslearn format + UTM projection) is completed.
* Adds `rslp.common.worker` system to launch workers that read tasks from a queue (Google Pub/Sub topic) and execute them. In the future all task systems (like Landsat and Sentinel-2 job launchers) should be converted to this system so that they can share the same way to execute workers -- then the individual projects just need to implement the process of writing tasks to the topic.


rslp/satlas/predict_pipeline.py
```diff
@@ -0,0 +1,368 @@
...
+        bounds: pixel coordinates within the projection on which to compute outputs.
+        time_range: time range to apply model on.
+        out_path: directory to write the outputs. It will either be a GeoTIFF or
+            GeoJSON, named based on the bounds.
+        scratch_path: where to store the dataset.
+        use_rtree_index: whether to prepare using rtree index. This is recommended when
+            applying the model globally but can be disabled for small regions to avoid
+            the time to create the index.
+    """
+    dataset_config_fname = DATASET_CONFIG_FNAME.format(application=application.value)
```
Hgherzog, line 106: Suggestion: Split this into a couple functions for readability

rslp/satlas/postprocess.py
```diff
@@ -0,0 +1,301 @@
...
+            labels.append(label)
+
+        # Sort by YYYY-MM, since the smoothing function expects us to provide all of
+        # the labels in temporal order.
+        labels.sort()
+
+        # Smoothing is handled by a Go script.
+        subprocess.check_call(
+            [
+                "rslp/satlas/scripts/smooth_point_labels_viterbi",
```
Hgherzog, line 272: Why can't the smooth point labels be written in python?

tests/integration_slow/satlas/test_predict_pipeline.py
```diff
@@ -0,0 +1,73 @@
...
+    # directory. Specifically, the scratch path will be populated with an rslearn
+    # dataset containing one window matching the projection/bounds that we provide.
+    out_path = tmp_path / "out"
+    scratch_path = tmp_path / "scratch"
+    out_path.mkdir()
+
+    # Apply the pipeline. It will ingest data and apply the model.
+    # We disable rtree index so that it doesn't need an hour to create it.
+    predict_pipeline(
+        application=Application.MARINE_INFRA,
```
Hgherzog, line 55: Would it make sense to have tests for the other applications as well?

rslp/launch_beaker.py
```diff
@@ -104,7 +100,9 @@ def launch_job(
                     config_path,
                     "--autoresume=true",
                 ],
-                constraints=Constraints(cluster=["ai2/jupiter-cirrascale-2"]),
+                constraints=Constraints(
+                    cluster=["ai2/jupiter-cirrascale-2", "ai2/augusta-google-1"]
```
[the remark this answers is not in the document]
favyen2: I will add refactoring of this to #100 since that PR consolidates some of the Beaker stuff.

rslp/satlas/data_sources.py
```diff
@@ -0,0 +1,393 @@
...
+
+    # If there are not enough matching mosaics, then we eliminate all the
+    # matches since we aren't going to use this window then anyway.
+    if len(cur_groups) < max_matches:
+        return []
+
+    return cur_groups
+
+
+class MonthlySentinel2(DataSource):
```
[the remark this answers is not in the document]
favyen2: Added tests. I think it makes sense for now to keep it in rslearn_projects, we will see how many applications use the same thing. If we do add to rslearn, we should refactor it so we don't need this wrapper class, ideally it'd be something in the QueryConfig maybe.

rslp/satlas/postprocess.py
favyen2: It may be feasible, I'm not sure, this is borrowed from https://github.com/allenai/remote-sensing-data-hub/blob/main/go_scripts/smooth_point_labels_viterbi.go and not much reason to rewrite at this time I think.

tests/integration_slow/satlas/test_predict_pipeline.py
favyen2: Currently it is only MARINE_INFRA / WIND_TURBINE and these are similar enough that I don't think we need to run both (the code is the same, just different model, but this test is just making sure the code runs correctly).

I think once we have SOLAR_FARM it could use its own test since there will be code specific to the segmentation.

rslp/launch_beaker.py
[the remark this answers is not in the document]
favyen2: In #100 I changed all the cluster options to be required so there are no default since different tasks will have different needs.

Hgherzog (approved):
Approved, though I mainly looked at how you addressed the comments. It is probably better to fix stuff as they come up at this point.

rslp/satlas/README.md
```diff
@@ -0,0 +1,124 @@
...
+  systems.
+- `gs://rslearn-eai/projects/satlas/marine_infra/version-20241212/2024-07/` is the
+  folder containing inference outputs that we want to merge.
+- `gs://rslearn-eai/projects/satlas/marine_infra/version-20241212/merged/` is the
+  folder to write merged outputs. The output filename will be
+  `gs://rslearn-eai/projects/satlas/marine_infra/version-20241212/merged/2024-07.geojson`.
+
+Second, smooth the points across timesteps. This runs a Viterbi smoothing operation.
+Note that the Viterbi smoothing is implemented in a separate Go application at
+`rslp/satlas/scripts/smooth_point_labels_viterbi.go`.
```
pbeukema, line 94: out of curiosity -- why did you decide to implement this with go (versus Python)

rslp/satlas/README.md
```diff
@@ -0,0 +1,124 @@
...
+Inference:
+
+    python -m rslp.main satlas write_jobs_for_year_months '[[2024, 1]]' WIND_TURBINE 'gs://rslearn-eai/projects/satlas/wind_turbine/version-20241210/{year:04d}-{month:02d}/' skylight-proto-1 rslp-job-queue-favyen --days_before 90 --days_after 181
+
+Post-processing:
+
+    python -m rslp.main satlas merge_points WIND_TURBINE 2024-01 gs://rslearn-eai/projects/satlas/wind_turbine/version-20241210/2024-01/ gs://rslearn-eai/projects/satlas/wind_turbine/version-20241210/merged/
+    python -m rslp.main satlas smooth_points WIND_TURBINE 2024-01 gs://rslearn-eai/projects/satlas/wind_turbine/version-20241210/merged/ gs://rslearn-eai/projects/satlas/wind_turbine/version-20241210/smoothed/
+
+Publishing for wind turbine is not supported yet since it needs to be combined with the
```
pbeukema, line 123: Where is this done? Is that "automated"?

favyen2: Not automated yet -- todos include adding the solar farm and tree cover models and then yeah scheduling it perhaps via Github action similar to the forest loss driver one.

So currently these steps would be executed manually. They can be executed from anywhere even small machine since the compute-intensive parts run on Beaker.

rslp/satlas/README.md
favyen2: It was a lot easier to get it to run efficiently in Go than in Python. In Go can just write code with loops while in Python have to deal with numpy operations over clusters that would have different number of members and such.

PR #96, opened 2025-01-31
Miscellaneous config changes.

favyen2:
- Use 32 workers when materializing vessel detection crops. This increases speed of the pipeline since otherwise it takes a long time just to crop the big image into the small crops.
- Just use Jupiter for Sentinel-2 job launcher. Other clusters have various issues. In the future this should use the `rslp.common.worker` system.
- Remove redundant code in Sentinel-2 vessel detection pipeline that could hide an error.
- Ecosystem mapping: update to use latest dataset version.
- Minderoo: removed unneeded inference config and update training config to new config format.


Hgherzog (approved):
LGTM

PR #97, opened 2025-02-04
Add GeoJSON output option for vessel detection pipelines

### LINKED ISSUE #95: Sentinel-2 / Landsat vessel detection pipeline improvements
* Make the crop directory so it doesn't fail at the end if it doesn't exist.
* Add GeoJSON output.
* Add visualization option (create one big PNG/GeoTIFF with rasterized vessels).

favyen2:
Add GeoJSON output option.

I also changed the order in NearInfraFilter to (longitude, latitude) to be consistent with rest of codebase.

And makes crop directory for Sentinel-2 vessel detection pipeline to avoid directory not exist error.

This depends on PR #96.

Resolves #95.

### DIFF
--- a/rslp/landsat_vessels/predict_pipeline.py
+++ b/rslp/landsat_vessels/predict_pipeline.py
@@ -304,12 +304,17 @@ def setup_dataset(
     if scene_zip_path:
         # Download the zip file and create image_files locally.
         scene_id = scene_zip_path.split("/")[-1].split(".")[0]
-        zip_dir = ds_path / "scene_zip"
+        zip_dir = (ds_path / "scene_zip").absolute()
+        zip_dir.mkdir(exist_ok=True)
         local_zip_path = os.path.join(zip_dir, scene_id + ".zip")
         download_and_unzip_scene(scene_zip_path, local_zip_path, zip_dir)
         image_files = {}
         for band in LANDSAT_BANDS:
-            image_files[band] = f"{zip_dir}/{scene_id}/{scene_id}_{band}.TIF"
+            # TODO: have helper utility function that gets str() of UPath with protocol
+            image_fname = str(zip_dir / scene_id / f"{scene_id}_{band}.TIF")
+            if "://" not in image_fname:
+                image_fname = f"file://{image_fname}"
+            image_files[band] = image_fname
 
     if image_files:
         # Setup the dataset configuration file with the provided image files.
--- a/rslp/landsat_vessels/predict_pipeline.py
+++ b/rslp/landsat_vessels/predict_pipeline.py
@@ -303,16 +303,13 @@ def setup_dataset(
 
     if scene_zip_path:
         # Download the zip file and create image_files locally.
-        tmp_zip_dir = tempfile.TemporaryDirectory()
-        zip_dir = tmp_zip_dir.name
         scene_id = scene_zip_path.split("/")[-1].split(".")[0]
+        zip_dir = ds_path / "scene_zip"
         local_zip_path = os.path.join(zip_dir, scene_id + ".zip")
         download_and_unzip_scene(scene_zip_path, local_zip_path, zip_dir)
         image_files = {}
         for band in LANDSAT_BANDS:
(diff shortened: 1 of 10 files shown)


rslp/landsat_vessels/predict_pipeline.py
```diff
@@ -430,31 +411,23 @@ def predict_pipeline(
...
+        if crop_window_dir is None:
             raise ValueError("Crop window directory is None")
         for band in ["B2", "B3", "B4", "B8"]:
             image_fname = (
-                detection.crop_window_dir
-                / "layers"
-                / LANDSAT_LAYER_NAME
-                / band
-                / "geotiff.tif"
+                crop_window_dir / "layers" / LANDSAT_LAYER_NAME / band / "geotiff.tif"
```
Hgherzog, line 430: Suggestion: This seems to be matching a conventional rslearn dataset format, if so we should use a function to format the directories

rslp/landsat_vessels/predict_pipeline.py
```diff
@@ -430,31 +411,23 @@ def predict_pipeline(
...
+        if near_infra_filter.should_filter(lon, lat):
             infra_detections += 1
             continue
         # Load crops from the window directory.
         images = {}
-        if detection.crop_window_dir is None:
+        crop_window_dir = detection.metadata["crop_window_dir"]
+        if crop_window_dir is None:
             raise ValueError("Crop window directory is None")
         for band in ["B2", "B3", "B4", "B8"]:
```
Hgherzog, line 428: Make these Bands into a constant so it is clear why they were chosen (or configurable based on the model if that makes sense here

rslp/sentinel2_vessels/job_launcher.py
```diff
@@ -61,10 +61,7 @@ def launch_job(image_name: str, tasks: list[PredictionTask]) -> None:
             ],
             constraints=Constraints(
                 cluster=[
-                    # Don't use Renton cirrascale since it loads very slowly from Weka.
-                    "ai2/prior-elanding",
                     "ai2/jupiter-cirrascale-2",
```
Hgherzog, line 64: Suggestion: Have a shared constant for clusters as I think this is used in multiple places

rslp/sentinel2_vessels/predict_pipeline.py
```diff
@@ -137,7 +119,7 @@ def get_vessel_detections(
         window.save_layer_datas(dict(SENTINEL2_LAYER_NAME=layer_data))
 
     logger.info("Materialize dataset for Sentinel-2 Vessel Detection")
-    apply_windows_args = ApplyWindowsArgs(group=group, workers=1)
+    apply_windows_args = ApplyWindowsArgs(group=group, workers=32)
     materialize_pipeline_args = MaterializePipelineArgs(
```
Hgherzog, line 123: Suggestion: Break this function out into smaller testable components

rslp/landsat_vessels/predict_pipeline.py
```diff
@@ -192,7 +171,7 @@ def run_classifier(
     for detection in detections:
```
Hgherzog, line 171: Suggestion: Brake this  into a couple different functions (perhaps it can share some of the functionality with the S2 pipeline)

rslp/sentinel2_vessels/predict_pipeline.py
Hgherzog: And add tests for each component

Hgherzog (approved):
Approved with some minor comments

rslp/landsat_vessels/predict_pipeline.py
favyen2: I think constant will be confusing since we reference each of these bands individually in the code below this, but I added comment explaining this section and what each of the bands is for.

rslp/sentinel2_vessels/job_launcher.py
favyen2: I added module-specific constant for now since the intention would be to replace all of these job launchers with the `rslp.common.worker` system (so each module just needs to write tasks to PubSub topic).

PR #98, opened 2025-02-05
Change --load_best=true to fail if checkpoint does not exist

### LINKED ISSUE #89: --load_best=true should force checkpoint to be loaded
Two issues:

- Currently even if `best.ckpt` or other checkpoint exists in the project folder, with `--load_best=true` the checkpoint will only be loaded if `last.ckpt` also exists since the first check is for `last.ckpt`. I think it should be updated so with `--load_best=true` it can load any checkpoint. If this is updated, then we can update the `docs/` to avoid needing to rename `best.ckpt` to `last.ckpt` when downloading checkpoints.
- It makes sense for `--autoresume=true` to continue even if there's no existing checkpoint, but I think `--load_best=true` should only continue if there's checkpoint instead of continuing with random initialization. Otherwise it is not obvious what the problem is when you are running `model test` or a prediction pipeline that it is really using random weights.

favyen2:
* Change --load_best=true to fail if checkpoint does not exist.
* Also adjust it to support loading checkpoint with arbitrary name in the checkpoint directory even if last.ckpt doesn't exist.

Resolves #89


tests/integration/test_lightning_cli.py
```diff
@@ -0,0 +1,25 @@
...
+import shutil
+from typing import Any
+
+import pytest
+from upath import UPath
+
+from rslp.utils.rslearn import run_model_predict
+
+
+def test_error_if_no_checkpoint(tmp_path: pathlib.Path, monkeypatch: Any) -> None:
```
Hgherzog, line 11: I would also add a test to ensure that the best checkpoint is indeed loaded if maybe multiple checkpoints exist

Hgherzog (approved):
Approved with minor comments

PR #100, opened 2025-02-06
Forest Loss Web Integration

favyen2:
- Change forest loss to run as a Beaker job, instead of spawning a single GCE VM which then spawns a single Beaker job. (The integrated pipeline in `rslp/forest_loss_driver/__init__.py` runs all those steps.)
- Store the intermediate rslearn dataset, along with tile store and index cache, on Weka. Tile store on Weka significantly accelerates steps like materialization because rasterio can do random reads from filesystem path or HTTP URL much faster than an arbitrary file object. The files are copied to GCS afterward.
- Move the beaker_launcher.py to `rslp/common/` since it was written to be task-agnostic anyway.
- Extend inference to a year instead of just 90 days.
- Materialize the Planet layers.
- Remove environment variables since there should not be so many places where configuration is stored.
- Remove `rslp/forest_loss_driver/inference/materialize_dataset.py` since it just passes the same arguments to another function.
- Remove `rslp/forest_loss_driver/inference/model_predict.py` since it just passes the same arguments to another function.
- Add steps to make the xyz tiles for display in forest-loss.allen.ai.
- Use tmp_path in tests instead of tempfile.TemporaryDirectory.

It is not working 100%, the Beaker job sometimes needs some restarting, but I think it is better to merge this set of changes first. Anyway the current forest loss deployment is no longer being used so I have to manually delete the stuff it creates before trying to run the new one.


Hgherzog (approved):
I may have missed some smaller issues but this looks good overall. If there are any other known issues (including the beaker restarting thing it would be good to document these in an issue

PR #105, opened 2025-02-14
Add API for Sentinel-2 vessel detection

favyen2:
Updates to the detector and also adding the API.


rslp/vessels/__init__.py
```diff
@@ -6,9 +6,38 @@
...
+        projection: the projection used.
+        score: confidence score from object detector.
+        ts: datetime fo the window (if known).
+        scene_id: the scene ID that the vessel was detected in (if known).
+        crop_fname: filename where crop image for this vessel is stored.
+        longitude: the longitude position of the vessel detection.
+        latitude: the latitude position of the vessel detection.
+    """
+
+    source: str
```
yawenzzzz, line 29: Change to StrEnum? I can also add this in a later PR.

yawenzzzz (approved):
LGTM! This actually reminds me to consolidate the docs on Landsat, and also update the prediction pipeline to use the new `VesselDetection`. I will do that in a later PR.

rslp/vessels/__init__.py
favyen2: thanks, str enum is a lot better here

PR #106, opened 2025-02-24
Run publish_project_docker_image.yaml reusable workflow from current branch

favyen2:
Run publish_project_docker_image.yaml reusable workflow from current branch.

Currently in master branch it is not working because for testing in previous PR I had it refer to the PR branch which is now deleted.

The change uses a different syntax that allows using reusable workflow in current branch (whichever one is executing the Action) rather than needing to refer to a specific branch.


yawenzzzz (approved):
LGTM

PR #113, opened 2025-04-08
Add code for launching Helios fine-tuning evaluations.

favyen2:
Add code for launching Helios fine-tuning evaluations.

There is documentation in `rslp/helios/README.md` but basically you can launch evaluation like this:

```
python -m rslp.main helios launch_finetune --helios_checkpoint_path /weka/dfive-default/helios/checkpoints/henryh/8latent_mim_random_patch_disc_new_exit_zero_lr_4e-05_base_shallow_decoder/step95000/ --experiment_prefix apr07test --image_name favyen/rslphelios --tasks '["eurosat"]'
```

Other changes:
- The positional arguments were getting out of hand so I changed it so arguments passed via command-line are all passed with --flag instead of positional. Most of the documentation changes are related to this.
- Moved rslp.launch_beaker to rslp.common.beaker_train so that it can use the workflow interface which means it doesn't need extra argparse code.
- There are also some fixes for Satlas pipeline to work with some changes in rslearn.

Todos:
- Try DETR object detection head since it may be better for transformer backbones.
- See if there are more classification tasks that would be good to evaluate on.

### DIFF
--- a/data/helios/satlas_wind_turbine/finetune.yaml
+++ b/data/helios/satlas_wind_turbine/finetune.yaml
@@ -26,7 +26,7 @@ model:
                 activation:
                   class_path: torch.nn.LayerNorm
                   init_args:
-                    normalized_shape: [192, 32, 32]
+                    normalized_shape: [192, {256/PATCH_SIZE}, {256/PATCH_SIZE}]
             - class_path: rslearn.models.conv.Conv
               init_args:
                 in_channels: 192
@@ -41,7 +41,7 @@ model:
                   class_path: torch.nn.Identity
             - class_path: rslearn.models.faster_rcnn.FasterRCNN
               init_args:
-                downsample_factors: [8]
+                downsample_factors: [{PATCH_SIZE}]
                 num_channels: 192
                 num_classes: 3
                 anchor_sizes: [[32]]
--- a/data/helios/satlas_wind_turbine/finetune.yaml
+++ b/data/helios/satlas_wind_turbine/finetune.yaml
@@ -1,0 +1,149 @@
+model:
+  class_path: rslearn.train.lightning_module.RslearnLightningModule
+  init_args:
+    model:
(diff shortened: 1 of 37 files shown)


data/helios/satlas_wind_turbine/finetune.yaml
```diff
@@ -0,0 +1,149 @@
...
+              checkpoint_path: "{CHECKPOINT_PATH}"
+              selector: ["encoder"]
+              forward_kwargs:
+                patch_size: 8
+        decoders:
+          detect:
+            - class_path: rslearn.models.conv.Conv
+              init_args:
+                in_channels: 768
+                out_channels: 192
```
yawenzzzz, line 19: Given that this `in_channels` is associated with a specific checkpoint, it may be better to make it configurable as `checkpoint_path`.

rslp/helios/launch_finetune.py
```diff
@@ -0,0 +1,94 @@
...
+                )
+
+                # I can't figure out how to override Helios checkpoint_path from
+                # command-line since it appears in a list, so instead we create a copy
+                # of the configuration file in a temporary directory.
+                with config_fname.open() as f:
+                    config_str = f.read()
+                config_str = config_str.replace(
+                    "{CHECKPOINT_PATH}", helios_checkpoint_path
+                )
```
yawenzzzz, line 64: see comment regarding `encoder_embedding_size`

yawenzzzz: also, `patch_size` could influence the performance, so making it configurable can also be good, see Table 12 in Galileo: https://arxiv.org/pdf/2502.09356, in Helios, the default `patch_size` is set to 4.

yawenzzzz:
LGTM! Just two minor comments.

yawenzzzz (approved):

PR #115, opened 2025-04-11
Crop type mapping

yawenzzzz:
1. Update Helios model & norm to support multitemporal inputs
2. Configs for IFRPI crop type mapping task: https://docs.google.com/document/d/1ZCtWWsdvhPRv6vHoGTtPN1FQSJguiubsYXXMkOjWUWc/edit?tab=t.0

Comparison between Helios (single timestep) + SimpleTimeSeries vs Helios (multiple timesteps):

https://wandb.ai/eai-ai2/helios_finetuning/runs/bb0p43j3
https://wandb.ai/eai-ai2/helios_finetuning/runs/pouenwn7


rslp/crop_type_mapping/create_windows.py
```diff
@@ -0,0 +1,244 @@
...
+    # Check if train or val.
+    if split_by_polygon:
+        group = "post_polygon_split"
+    else:
+        group = "post_random_split"
+    window_name = f"{polygon_id}_{latitude}_{longitude}"
+
+    # If split by polygon id, no samples from the same polygon will be in the same split.
+    if split_by_polygon:
+        is_val = hashlib.md5(str(polygon_id).encode()).hexdigest()[0] in ["0", "1"]
```
favyen2, line 127: I have found that sometimes MD5 can give poor hashes (end up with biased distributions), would recommend sha256 for the future. I think it's fine to keep it as-is for this PR if desired since it may be troublesome to re-generate the windows.

favyen2 (approved):

rslp/crop_type_mapping/create_windows.py
yawenzzzz: I see, thanks! I'm generating datasets for WorldCereal, I will use sha256 for this task then.

PR #119, opened 2025-04-24
fix docs for landsat model versions

favyen2:
There were a couple missing versions in between that did not get documented.


yawenzzzz (approved):
LGTM!

PR #121, opened 2025-04-25
add prom metrics to sentinel2

hunterp:
Adds prometheus timers to the sentinel2 vessel detection API.


rslp/landsat_vessels/prom_metrics.py
```diff
@@ -6,7 +6,10 @@
 from prometheus_client.context_managers import Timer
 
 request_timer = Histogram(
-    "landsat_rslearn_timer", "Timers for inference requests", ["operation"]
+    "landsat_rslearn_timer", "Timers for inference requests",
+    labelnames=["operation"],
+    buckets=[0.01, 0.05, 0.1, 0.5, 1, 2, 5, 10, 30, 60, 120, (60 * 3), (60 * 4), (60 * 5), (60 * 6), (60 * 7), (60 * 8),
```
hunterp, line 11: default prometheus histograms are optimized for timing HTTP requests; which are typically quite fast.  the highest bucket is 10 seconds; which we're significantly exceeding here.  These buckets hopefully get us better visibility into some of the longer running operations.

favyen2 (approved):

PR #129, opened 2025-05-09
Add CROMA/Helios fine-tuning configs for team meeting presentation.

favyen2:
Add CROMA/Helios fine-tuning configs for team meeting presentation

I also added embedding_size/patch_size arguments to Helios so that it can return the right values for get_backbone_channels. This is used by the SimpleTimeSeries module (which is needed for comparing the bitemporal images for forest loss driver classification).


yawenzzzz (approved):
LGTM!

PR #132, opened 2025-05-13
Add debug detr config

favyen2:
It is example of applying DETR for some object detection task.


favyen2:
Hm I think I opened #134 and forgot about this PR, I will close this one since that one is more relevant.

PR #138, opened 2025-05-27
Use Beaker queues instead of Google Cloud Pub/Sub for the worker system.

favyen2:
Main changes:
- Use Beaker queues instead of Google Cloud Pub/Sub for the worker system
- Upgrade beaker-py to 2.x (queues are not supported in older version).

Minor updates/fixes:
- Update `helios.Dockerfile` to use pytorch 2.7.0 to reduce build time.
- Update `one_off_projects/convert_satlas_webmercator_to_rslearn/lib/__init__.py` for new rslearn VectorFormat.encode_vector API.
- Remove `manage_scratch_dir_on_data_disk` option since it's not needed anymore (since the Docker volumes are now on the big `/data` disk across all Beaker nodes).


yawenzzzz (approved):
LGTM!

PR #141, opened 2025-05-27
Add missing caching functionality from Satlas data sources.

favyen2:
We should move the monthly functionality to rslearn so this maintenance isn't needed. It is on my TODO list but I haven't gotten around to it yet, so for now still have this PR.


favyen2:
I merged this into #137 since that PR affects the same file.

PR #144, opened 2025-06-03
Support specifying retries on the Beaker job

favyen2:
Support specifying retries on the Beaker job

Depends on #138


yawenzzzz (approved):
LGTM!

PR #145, opened 2025-06-09
Add time series and multimodal configs for marine_infra and solar_farm

favyen2:
Add time series and multimodal configs for marine_infra and solar_farm.

I also updated the launcher code to accept specifying a list of configs. This enables reducing duplication between some of the config files, although makes it more complicated to start the experiments since you need to specify a list of configs to get the right combination (this is documented in the README files within each task dir in data/helios though).


yawenzzzz (approved):
LGTM! I think this base config & overwrites is nice, but I'm thinking could we also separate by data / model / trainer, but the data part may involve different transformations per model. I will try it out, see which one works more convenient.

PR #153, opened 2025-06-26
Add CROMA comparison for the Helios fine-tuning experiments

favyen2:
Add CROMA comparison for the Helios fine-tuning experiments.

I also increased batch size for Helios for PASTIS and marine_infra, since it uses less memory now with flash attention.


yawenzzzz (approved):
LGTM!

PR #154, opened 2025-06-26
Save config.json to the project directory in RSLP_PREFIX

favyen2:
Save config.json to the project directory in RSLP_PREFIX.

I also fixed bugs when RSLP_PREFIX is local filesystem relating to needing to create some parent directories. This way you can set RSLP_PREFIX to /weka/dfive-default/rslearn-eai/ before running `model fit`; it also works when launching job with `common beaker_train` since it takes the RSLP_PREFIX from the environment and passes it on to the Beaker job. Note that you will still need to pass command-line option to enable corresponding WEKA mount.


rslp/lightning_cli.py
```diff
@@ -111,6 +114,33 @@ def on_fit_start(self, trainer: Trainer, pl_module: LightningModule) -> None:
             wandb.config.update(json.loads(self.config_str))
 
 
+class SaveConfigToProjectDirCallback(SaveConfigCallback):
+    """Callback to save the configuration to checkpoint directory."""
+
+    def save_config(
```
Hgherzog, line 120: nit: Add a comment that lightning is implicitly calling this on rank zero only to prevent a race condition.

Hgherzog (approved):
LGTM

PR #165, opened 2025-07-11
Finetuning on concatenated datasets

ryspark:
Mostly just applying changes from rslearn (https://github.com/allenai/rslearn/pull/207)
* Adds support for finetuning on multiple datasets in memory at once (though only one task at a time)
* Adds support for parameter-efficient finetuning via [ALPA](https://arxiv.org/html/2503.11335v2), but this doesn't give that much of a performance boost and is not recommended

New configs in `data/helios/v3_multitask` and `data/helios/v3_perf_benchmark` are for multi-dataset training. Some of them are just for debugging and will be cleaned up in a future PR.


data/helios/v2_landsat_vessels/finetune_detector_startv1.yaml
```diff
@@ -0,0 +1,138 @@
+model:
```
favyen2, line 1: If the purpose for most of these new model configs is for the multi-task / cross-task fine-tuning experiments, I think they should go in a new folder specifically for that like `one_off_projects/2025_07_multitask_finetuning`.

rslp/helios/launch_finetune.py
```diff
@@ -6,105 +6,222 @@
...
 
 def launch_finetune(
-    helios_checkpoint_path: str,
     experiment_id: str,
-    image_name: str,
-    encoder_embedding_size: int,
-    patch_size: int,
-    cluster: list[str],
-    config_paths: list[str],
+    config_paths: "list[str]",
```
favyen2, line 24: Elsewhere we use `list[str]` instead of `"list[str]"`.

rslp/helios/launch_finetune.py
```diff
@@ -6,105 +6,222 @@
...
+                    model_path = config["model"]["init_args"]["model"]["init_args"][
+                        "checkpoint_path"
+                    ]
+                    model_path_parts = model_path.split(os.path.sep)
+                    model_name = (
+                        model_path_parts[-3]
+                        + "_"
+                        + model_path_parts[-1].replace(".ckpt", "")
+                    )
+                eval_task = "__".join(config_paths[0].split(os.path.sep)[-2:]).strip(
```
favyen2, line 117: Is it possible to reuse the rslp_project and rslp_experiment here instead of extracting a model name and task name from the provided files?

rslp/helios/launch_finetune.py
```diff
@@ -6,105 +6,222 @@
 import tempfile
 from pathlib import Path
 
+import yaml
+
 from rslp.log_utils import get_logger
 
+DEFAULT_RSLP_PREFIX = "project_data/"
```
favyen2, line 13: I don't think this script should have defaults for RSLP_PREFIX, since the assumption everywhere else in this codebase is that the user is responsible for setting the environment variable. It can be placed in .env file.

rslp/helios/model.py
```diff
@@ -83,7 +85,27 @@ def __init__(
...
         if not random_initialization:
             train_module_dir = f"{checkpoint_path}/model_and_optim"
-            load_model_and_optim_state(train_module_dir, model)
+            if os.path.exists(train_module_dir):
+                load_model_and_optim_state(train_module_dir, model)
+                print(
+                    f"INFO: loaded helios encoder from {train_module_dir}/model_and_optim"
+                )
+
+            else:
```
favyen2, line 94: I think it would be better to handle this code path by passing the RestoreConfig directly to RslearnLightningModule.

The checkpoint path option is only provided here because the Helios checkpoints need to be handled in a different way (due to the distributed saving).

rslp/lightning_cli.py
```diff
@@ -331,6 +336,26 @@ def before_instantiate_classes(self) -> None:
...
+            # and so don't need lightning to wrap it for us
+            if c.data.class_path == "rslearn.train.data_module.MultiDatasetDataModule":
+                # Usually, it's fine to leave this flag on (lightning will detect
+                # a distributed sampler and leave it alone), but since we have a custom
+                # one not subclassing DistributedSampler, we need to turn it off manually
+                logger.info("Using custom distributed sampler")
+                logger.info(
+                    "Warnings about calling compute on an empty set of metrics may appear"
+                )
+                c.trainer.use_distributed_sampler = False
```
favyen2, line 357: This should already always be disabled here (recent change):
https://github.com/allenai/rslearn/blob/8141e1d60e78265828ad3c920953fca160e10338/rslearn/main.py#L807
So this code may not be needed.

tests/unit/maldives_ecosystem_mapping/test_train.py
```diff
@@ -10,10 +12,8 @@ class TestModel(torch.nn.Module):
     def __init__(self) -> None:
         super().__init__()
 
-    def forward(
-        self, inputs: torch.Tensor, targets: torch.Tensor
-    ) -> tuple[torch.Tensor, dict]:
-        return inputs, {"loss": 0}
+    def forward(self, inputs: torch.Tensor, targets: torch.Tensor) -> dict[str, Any]:
+        return {"outputs": inputs, "loss_dict": {"loss": 0}}
```
favyen2, line 16: Does this depend on changes to rslearn that have not been merged yet?

favyen2 (approved):
This looks good to me, I just left one comment about the test change to return dict that I did not follow.

PR #167, opened 2025-07-14
Kenya/Nandi crop type inference

yawenzzzz:
An initial idea is to split the whole county into 4K windows (each is 128*128 pixel), and perform inference on each window. Here, the approach is to create patches per window, and then run model forward on all the patches. 

Command for local run: `python -m rslp.rslearn_main model predict --config data/helios/v2_nandi_crop_type/finetune_s2.yaml`


yawenzzzz:
The inference pipeline has been updated #170

PR #169, opened 2025-07-24
Add esrunner sample project with how-to guide.

joshhvulcan:


data/esrunner_sample/README.md
```diff
@@ -0,0 +1,171 @@
+# ES Runner Local Development Guide
+
+## What is esrunner?
+
+ESRunner provides the [EsPredictRunner](https://github.com/allenai/earth-system-run/blob/josh/esrunner/src/esrun/runner/local/predict_runner.py) (and perhaps eventually the EsFineTuneRunner) class which can be used to run predictions and fine-tuning pipelines outside of the esrun service. 
+
+## Setting up your environment
+
+- Install `esrunner` (earth-system-run) in your development environment (Or clone the repository and add to your `PYTHONPATH`. If you go this route, ensure you install the packages listed in `earth-system-run/requirements.txt`)
```
hunterp, line 9: I do not think anybody should clone the repo.

data/esrunner_sample/README.md
```diff
@@ -0,0 +1,171 @@
...
+- Following the project structure below, create a directory in the `rslearn-projects/data/` directory. This directory will contain all the necessary files for your prediction or fine-tuning pipeline.
+
+## Project Structure
+- `checkpoint.ckpt`: (Optional)
+- `dataset.json`: This is the rslearn dataset definition file.
+- `model.yaml`: This is the rslearn (pytorch) model definition file.
+- `partition_strategies.yaml`: 
+- `postprocessing_strategies.yaml`: This file defines how the esrunner will post-process the predictions.  
+- `requirements.txt`: This file contains the additional Python packages required for the pipeline. It should include any dependencies that are not part of the base environment.
+- `prediction/test-request1.geojson`: This directory contains the prediction requests in GeoJSON format. Each file represents a set of prediction requests for a specific region or time period.  Many different prediction requests can be defined within a single file as separate features in the feature collection. The esrunner will partition these requests into smaller tasks based on the partition strategies defined in `partition_strategies.yaml`.
```
hunterp, line 19: why is this a directory of geojson files  vs just a single feature collection?

data/esrunner_sample/README.md
```diff
@@ -0,0 +1,171 @@
...
+- `run_pipeline.py`: This script is used to run the prediction pipeline. It will read the configuration files and execute the necessary steps to perform predictions or fine-tuning. You can customize this script to suit your specific needs, such as adding additional logging or error handling.
+
+## Partitioning Strategies
+This file defines how the esrunner will partition inference request into compute tasks (large) and prediction windows (small).  Each large partition is equivalent to an rslearn window group and contains multiple small partitions.  The small partitions are the individual rslearn prediction windows.
```
hunterp, line 23: please replace this with the new names from Sebastian. 

```
partition_large -> partition_request_geometry
partition_small -> prepare_window_geometries
```

data/esrunner_sample/README.md
```diff
@@ -0,0 +1,171 @@
+# ES Runner Local Development Guide
+
+## What is esrunner?
+
+ESRunner provides the [EsPredictRunner](https://github.com/allenai/earth-system-run/blob/josh/esrunner/src/esrun/runner/local/predict_runner.py) (and perhaps eventually the EsFineTuneRunner) class which can be used to run predictions and fine-tuning pipelines outside of the esrun service. 
+
+## Setting up your environment
+
+- Install `esrunner` (earth-system-run) in your development environment (Or clone the repository and add to your `PYTHONPATH`. If you go this route, ensure you install the packages listed in `earth-system-run/requirements.txt`)
+- Following the project structure below, create a directory in the `rslearn-projects/data/` directory. This directory will contain all the necessary files for your prediction or fine-tuning pipeline.
```
hunterp, line 10: if we're telling folks to have the `rslp/data/{my_project}/` directory match this structure; why do we need to have each filename passed in?

data/esrunner_sample/README.md
```diff
@@ -0,0 +1,171 @@
...
+  class_path: esrun.tools.partitioners.fixed_window_partitioner.FixedWindowPartitioner
+  init_args:
+    window_size: 128 # intended to be a pixel value
+```
+
+## Post-Processing Strategies
+There are 3 different stages to postprocessing:
+  - window (small) - This is the stage where outputs from the model are converted into a digestible artifact for the next stage.
+  - partition (medium) - This is the stage where the predictions for each small partition are combined into a single per-partition artifact.
+  - dataset (large) - This is the final stage of postprocessing where the predictions are combined into a artifact.
```
hunterp, line 51: I think you actually just came up with the better names here :laughing: 

```
postprocess_window
postprocess_partition
postprocess_dataset
```

data/esrunner_sample/README.md
```diff
@@ -0,0 +1,171 @@
...
+Ideally the run_pipeline.py script could be replaced with a cli tool that automatically saves progress and state and allows you to re-run specific steps instead of the whole pipeline. eg:
+
+```bash
+esrunner --local \
+         --dataset-config /path/to/dataset.json \
+         ... \
+         --run inference \
+         --partition-id my-partition-id
+```
+But this is a longer term goal and not currently implemented.
```
hunterp, line 171: I would remove this before checking in.

pyproject.toml
```diff
@@ -21,9 +21,12 @@ build-backend = "setuptools.build_meta"
...
 include = ["rslp*"]
 
-[tool.setuptools.dynamic]
-dependencies = {file = ["requirements.txt"]}
-optional-dependencies = {ai2 = { file = ["ai2_requirements.txt"] }}
+[tool.setuptools.dynamic.dependencies]
+file = ["requirements.txt"]
+
+[tool.setuptools.dynamic.optional-dependencies]
+ai2 = { file = ["ai2_requirements.txt"] }
```
hunterp, line 28: I don't see ai2_requirements.txt here.  

Can we chat with @StephenWithPH about the path forward here as well.

joshhvulcan: it already existed.  I'm just wiring it up to be accessible via `pip install rslearn_projects[ai2]`.  Totally agree on the Stephen thing.

data/esrunner_sample/README.md
joshhvulcan: This is one of those things I want to validate with the ML folks. There are _a lot_ of cases where they store a variety of model configs for different experiments in the same directory. I am assuming they will want to continue doing that to some degree.  Perhaps a happy medium would be to read the prescribed names but also allow them to be overridden in the `EsPredictionRunner.__init__()` for flexibility.

data/esrunner_sample/README.md
joshhvulcan: This came from a conversation with Henry where he mentioned wanting to be able to work with different input geometries.  I figured providing a pattern for managing these different inputs would work better than saying "you must only have one input file".

data/esrunner_sample/README.md
cmwilhelm: Could it not be part of the project-specific requirements.txt, pointing to a git repo at a specific sha? that way when we're installing deps for the docker image we can reliably just say `pip install rslp-projects/thingy/requirements.txt` (or requirements.frozen.txt).

They should _probably_ be running a separate venv per project, so this doesn't seem like a hard sell?

favyen2: Per our conversation toady, we intend to continue to have one set of requirements for all of `rslearn_projects`. We use one environment and build one Docker container for all projects. Projects (as in different applications for which we want to fine tune models) share the vast majority of requirements and those that are not shared are typically specific to individual model architectures like TerraMind vs OLMo-Earth so even then they would not be project-specific, unless "experiments to compare OLMo-Earth against TerraMind / DINOv2" is considered a project which is not really how we think about it. There are some requirements like prometheus-client that are only used for specific projects like vessel detection but I think the intention is to make things more consistent across projects e.g. using the same system for observability.

joshhvulcan:
@favyen2 I am probably going to close this one and reimplement from master so that its all clean. A lot has changed since I opened this so its probably best just to start fresh and build back up.  I will put that on my list for today.

favyen2:
Sounds good maybe can have separate PRs for adding the example versus building the Docker container for esrun (which may involve updates to how dependencies are split up).

PR #170, opened 2025-07-29
Update Kenya Nandi county crop type project to use segmentation

favyen2:
This depends on changes in https://github.com/allenai/rslearn/pull/223

Remaining things to do:
* Use the padding from the RasterMerger. The main benefit is that the predictions won't be offset (currently the prediction at each 4x4 input shows up in the output at the topleft pixel of that 4x4, but really it should be more centered), since anyway the underlying model outputs the same class at each pixel.

### DIFF
--- a/rslp/crop/kenya_nandi/train.py
+++ b/rslp/crop/kenya_nandi/train.py
@@ -3,7 +3,6 @@
 from typing import Any
 
 import torch
-import torch.nn.functional as F
 from rslearn.models.pooling_decoder import PoolingDecoder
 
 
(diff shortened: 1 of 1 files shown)


rslp/crop/kenya_nandi/train.py
```diff
@@ -0,0 +1,69 @@
...
+
+        This only works when all of the pixels have the same segmentation target.
+        """
+        output_probs = super().forward(features, inputs)
+        # BC -> BCHW
+        h, w = inputs[0]["sentinel2_l2a"].shape[1:3]
+        return output_probs[:, :, None, None].repeat([1, 1, h, w])
+
+
+class MultiSegmentationPoolingDecoder(torch.nn.Module):
```
yawenzzzz, line 26: What is this one used for?

PR #174, opened 2025-08-05
Add alphaearth supplementary eval datasets

favyen2:
Create rslearn datasets with Sentinel-2 and other satellite images from the AlphaEarth supplementary evals.

A script to save Helios embeddings, along with a script to compute balanced accuracy for k-nearest-neighbor, are also added.


one_off_projects/2025_08_01_alphaearth_eval_to_rslearn/config.json
```diff
@@ -0,0 +1,254 @@
...
+      ],
+      "data_source": {
+        "gcs_bucket_name": "rslearn-gee-ingest-bucket",
+        "index_cache_dir": "cache/gse",
+        "ingest": false,
+        "name": "rslearn.data_sources.google_earth_engine.GoogleSatelliteEmbeddings",
+        "service_account_credentials": "/etc/credentials/gee_credentials.json",
+        "service_account_name": "helios-dataset-creation",
+        "query_config": {
+          "max_matches": 8
```
yawenzzzz, line 247: Given that the support time span is no more than 1 year, `max_matches = 2` should be sufficient. I'm also wondering why some of the embeddings are all zero?

rslp/helios/get_embeddings.py
```diff
@@ -0,0 +1,205 @@
...
+            ]
+            features = helios(gpu_inputs)[0]
+
+            for cur_feat, metadata in zip(features, metadatas):
+                if args.mode == "center":
+                    selected = cur_feat.cpu().numpy()[
+                        :, cur_feat.shape[1] // 2, cur_feat.shape[2] // 2
+                    ]
+                elif args.mode == "pool":
+                    selected = cur_feat.cpu().numpy().mean(axis=(1, 2))
```
yawenzzzz, line 195: The results here: https://docs.google.com/spreadsheets/d/12SbIcQUWS5sivgXHiMKqFz_5_2Yg6YXy6cfly9cSYRk/edit?gid=1476525267#gid=1476525267, are they using center or pool? Also, the center model seems more accurate with small patch sizes (1–2), for larger patches it can be offset since the input size is even.

rslp/helios/get_embeddings.py
```diff
@@ -0,0 +1,205 @@
...
+                else:
+                    raise ValueError(f"invalid mode {args.mode}")
+                out_fname = (
+                    dataset.path
+                    / "windows"
+                    / metadata["group"]
+                    / metadata["window_name"]
+                    / args.embed_fname
+                )
+                np.save(out_fname.path, selected)
```
yawenzzzz, line 205: Suggestion: Split the main logic into functions (e.g., build dataset, load model, get embeddings). This makes it easier to reuse and extend, e.g., adding function to visualize embeddings for analysis.

one_off_projects/2025_08_01_alphaearth_eval_to_rslearn/get_balanced_accuracy.py
```diff
@@ -0,0 +1,233 @@
...
+) -> torch.Tensor:
+    device = torch.device("cuda")
+    train_embeddings = train_embeddings.to(device=device)
+    train_labels = train_labels.to(device=device)
+    test_embeddings = test_embeddings.to(device=device)
+
+    cos = nn.CosineSimilarity(dim=-1)
+    all_preds = []
+    for idx in range(test_embeddings.shape[0]):
+        test_embedding = test_embeddings[idx].unsqueeze(dim=0)
```
yawenzzzz, line 83: This can be deleted as `test_embedding` is assigned below.

one_off_projects/2025_08_01_alphaearth_eval_to_rslearn/get_balanced_accuracy.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def load_location_embedding(window: Window) -> torch.Tensor:
+    wgs84_geom = window.get_geometry().to_projection(WGS84_PROJECTION)
+    lon = wgs84_geom.shp.centroid.x
+    lat = wgs84_geom.shp.centroid.y
+    return torch.tensor([lon, lat])
+
+
+def load_pixel_embedding(window: Window) -> torch.Tensor:
```
yawenzzzz, line 44: return `torch.Tensor | None`

yawenzzzz (approved):

one_off_projects/2025_08_01_alphaearth_eval_to_rslearn/config.json
favyen2: I think the embeddings being zero happens when there is a mismatch between the reported bounds of the image and the actual data in that image. So e.g. if GEE reports a bounding box of the image but there are some portions that have no data.

For the same reason, I think max_matches=8 is safer since there could be multiple images at the same timestep that contain the window based on the reported geometry but the actual image doesn't intersect it.

I think it could be possible to use the compositing code that Rohan added to rslearn so it creates one composite instead of needing to create multiple mosaics. Composite here basically means it will allow for multiple items to be put in the same item group even if they don't add anything that wasn't covered by other items.

PR #180, opened 2025-08-16
West Kenya crop type segmentation

yawenzzzz:


yawenzzzz:
Close this one as we paused the work on JRC dataset, they will use either our released weights (as they already using Galileo) or ES run to do this task.

PR #188, opened 2025-08-19
Nandi crop type inference with esrun

yawenzzzz:
1. Added the model configs for nandi, updated the documentation
2. Updated the model config for mangrove to support inference.


esrun_data/crop/nandi/model.yaml
```diff
@@ -28,7 +32,7 @@ model:
 data:
   class_path: rslearn.train.data_module.RslearnDataModule
   init_args:
-    path: /weka/dfive-default/rslearn-eai/datasets/crop/kenya_nandi/20250616/
+    path: ${DATASET_PATH}
```
gabrieltseng, line 35: @joshhvulcan where / how are these env vars defined?

gabrieltseng: found it

esrun_data/nandi/model.yaml
```diff
@@ -0,0 +1,321 @@
+# FIXME ${TRAINER_DATA_PATH}
```
favyen2, line 1: You should be able to set these similar to https://github.com/allenai/rslearn_projects/blob/master/esrun_data/satlas/solar_farm_oe/model.yaml

esrun_data/nandi/model.yaml
```diff
@@ -0,0 +1,321 @@
...
+model:
+  class_path: rslearn.train.lightning_module.RslearnLightningModule
+  init_args:
+    model:
+      class_path: rslearn.models.multitask.MultiTaskModel
+      init_args:
+        encoder:
+          - class_path: rslp.helios.model.Helios
+            init_args:
+              checkpoint_path: ${EXTRA_FILES_PATH}/helios/step300000
```
favyen2, line 14: Document what EXTRA_FILES_PATH should be somewhere until we can pass it in esrun.yaml.

favyen2 (approved):
I didn't look carefully at the renamed files but overall looks good to me

esrun_data/nandi/model.yaml
yawenzzzz: Added into readme.

yawenzzzz:
> I didn't look carefully at the renamed files but overall looks good to me

Yeah, I fixed the names in most of the places, there maybe missed one, I can add a fix later. But also @gabrieltseng The mozambique PR will need one small change in the model config: from `rslp.crop.kenya_nandi.train.SegmentationPoolingDecoder` to `rslp.nandi.train.SegmentationPoolingDecoder`

PR #190, opened 2025-08-19
Fix components of LFMC workflow that were not open

favyen2:
Fix components of LFMC workflow that were not open, and include instructions for performing inference over a new region.

One thing I am not sure about is I ran the pipeline for the new coordinates in the README (37.19, -121.53) at 2025-03 and 2025-08 but the average LFMC over those 32x32 windows only changes from 101 to 82. Which is the expected trend but seems less dramatic than should be.

Depends on a minor change https://github.com/allenai/rslearn/pull/240


rslp/lfmc/README.md
```diff
@@ -1,51 +1,96 @@
...
+
 ### 2. Create Windows
 
-Run the following commands to create windows:
+Run the following commands to create windows (note that this script requires pandas to
+be additionally installed):
 ```
-python -m rslp.lfmc.create_windows --csv_path /tmp/lfmc-labels.csv --ds_path /weka/dfive-default/rslearn-eai/datasets/lfmc/20250626 --window_size 32
+export DATASET_PATH=/path/to/dataset
+export DATASET_GROUP=global_lfmc
```
APatrickJ, line 30: I know you copied this over, but let's use globe_lfmc for consistency with the original dataset name.

favyen2:
I think you need to copy the data/lfmc/config.json to the dataset path. I will update the tar file to include updated config.json. The one in the tar file doesn't include the output layer so you get this vague error `'output'` ._. jsonargparse errors are not very good, maybe there is a way to tell it to not catch the exception when instantiating class.

APatrickJ (approved):

PR #191, opened 2025-08-21
Add LFMC site name splits

APatrickJ:
This updates future invocations to use site name for splits and adds a standalone script that can be run to update the splits for an existing dataset.


rslp/lfmc/assign_site_name_splits.py
```diff
@@ -0,0 +1,32 @@
+"""Assign split based on site name."""
```
favyen2, line 1: Suggest to include documentation here that this script exists because the windows were originally created without "site_name_split".

Actually it may be better to move this script to one_off_projects since we shouldn't need to maintain it after running it once with `create_windows.py` also updated.

favyen2 (approved):

one_off_projects/2025_10_02_lfmc/assign_site_name_splits.py
```diff
@@ -0,0 +1,32 @@
+"""Assign split based on site name."""
```
APatrickJ: Added a README and moved the script to `one_off_projects`. Good idea.

favyen2 (approved):

PR #199, opened 2025-08-29
Try to get fine-tuned olmoearth for solar farm on esrun

favyen2:


joshhvulcan (approved):
Excellent! I will try the oe version in esrun!

PR #201, opened 2025-09-04
Use UPath in helios/model.py

joshhvulcan:


joshhvulcan:
@favyen2 Updated to include `.path` at your suggestion. I've been looking through `load_model_and_optim_state()` and it appears to support reading remote files so I didn't add a local file check.

favyen2:
> @favyen2 Updated to include `.path` at your suggestion. I've been looking through `load_model_and_optim_state()` and it appears to support reading remote files so I didn't add a local file check.

Hm OK I was looking at https://github.com/allenai/OLMo-core/blob/main/src/olmo_core/io.py#L40 but I guess it may still be able to handle some protocols. Sorry in that case maybe it's better to pass the UPath or str(path) lol; because x.path will just include the path component not the protocol.

joshhvulcan:
> Hm OK I was looking at https://github.com/allenai/OLMo-core/blob/main/src/olmo_core/io.py#L40 but I guess it may still be able to handle some protocols. Sorry in that case maybe it's better to pass the UPath or str(path) lol; because x.path will just include the path component not the protocol.

I think that just strips off the `file://`, but will leave `gs://` in place.  `load_model_and_optim_state()` uses `RemoteFileSystemReader` to read the file which eventually drops into [this block](https://github.com/allenai/OLMo-core/blob/main/src/olmo_core/io.py#L127-L130) in `get_bytes_range()` which operates on the `gs` protocol.  Anyway, I'm just going to pass the `str()` version through since that seems the safest at this point.

PR #204, opened 2025-09-10
Galileo config for pastis

gabrieltseng:
Depends on https://github.com/allenai/rslearn/pull/272


favyen2 (approved):

robmarkcole:
Is this just using the dataset as is from https://zenodo.org/records/5012942 ?

favyen2:
> Is this just using the dataset as is from https://zenodo.org/records/5012942 ? I guess not, in which case it would be nice to know how it is generated, presumably with rslearn

Here is the script used to convert the data:

https://github.com/allenai/rslearn_projects/blob/master/one_off_projects/2025_05_13_pastis/convert.py

It is just converted to rslearn dataset format, but the images still come from the original dataset (not generated by rslearn).

robmarkcole:
@gabrieltseng it is also necessary to include the required `config.json` thanks

gabrieltseng:
@robmarkcole I have been using GPUs with less memory than a H100 for testing. It may be worth increasing the patch size or decreasing the model size if you are running into memory issues

PR #218, opened 2025-10-03
Latest esrun configs

joshhvulcan:


favyen2:
I don't understand the reason for these changes, it seems to be reverting all of the changes I made to get it working with the latest esrun.

PR #219, opened 2025-10-03
LFMC on OlmoEarth Run

APatrickJ:
**If you are looking here, please check #230 for the updated configs**

### DIFF
--- a/olmoearth_run_data/lfmc/README.md
+++ b/olmoearth_run_data/lfmc/README.md
@@ -6,7 +6,7 @@ export DATASET_PATH=/weka/dfive-default/rslearn-eai/datasets/lfmc/20251007-scrat
 export NUM_WORKERS=32
 export TRAINER_DATA_PATH=/weka/dfive-default/patrickj/test/lfmc
 export WANDB_PROJECT=2025_10_08_lfmc
-export WANDB_NAME=lfmc_helios_base_s2_ts_ws1_ps1
+export WANDB_NAME=lfmc_helios_base_s1_s2
 export WANDB_ENTITY=eai-ai2
 ```
 


rslp/esrun/tools/data_splitters/spatial_data_splitter.py
```diff
@@ -0,0 +1,93 @@
...
+        utm_centroid = cast(BaseGeometry, utm_geometry.shp).centroid
+
+        # Calculate UTM tile coordinates
+        utm_tile = (
+            int(utm_centroid.x // self.grid_size),
+            int(utm_centroid.y // self.grid_size),
+        )
+
+        # Create deterministic hash from tile coordinates
+        tile_str = f"{utm_tile[0]}_{utm_tile[1]}"
```
favyen2, line 80: Should this include the UTM CRS? Otherwise all UTM zones will have the same tiling within the zone, which may or may not be desirable (I think not desirable but not sure).

esrun_data/lfmc/model.yaml
```diff
@@ -0,0 +1,174 @@
...
+        dtype: FLOAT32
+        load_all_item_groups: true
+        load_all_layers: true
+      sentinel2_l2a:
+        data_type: "raster"
+        layers: ["sentinel2"]
+        bands: ["B02", "B03", "B04", "B08", "B05", "B06", "B07", "B8A", "B11", "B12", "B01", "B09"]
+        passthrough: true
+        dtype: FLOAT32
+        load_all_item_groups: true
```
favyen2, line 63: This may be fine but currently there's no way to say you want minimum X timesteps for each example (and also I think the number of examples needs to be consistent across examples) so if you have some examples with less than the expected number of timesteps (no images captured there for part of the year for whatever reason) then it may cause an error. The solution for now is to use `load_all_layers: true` but not `load_all_item_groups`, instead list all those groups within the `layers` (`layers: ["sentinel2", "sentinel2.1", "sentinel2.2", "sentinel2.3", etc.]`).

esrun_data/lfmc/model.yaml
```diff
@@ -0,0 +1,174 @@
...
+    - class_path: rslearn.train.prediction_writer.RslearnWriter
+      init_args:
+        path: /weka/dfive-default/rslearn-eai/datasets/crop/kenya_nandi/20250616/
+        output_layer: prediction_v4
+        selector: ["crop_type_classification"]
+        merger:
+          class_path: rslearn.train.prediction_writer.RasterMerger
+          init_args:
+            padding: 2
+rslp_project: 2025_10_02_lfmc
```
favyen2, line 173: Comment these out for esrun (or remove them entirely if you are fine-tuning within esrun; reason to keep it would be so it's clear where the checkpoint is).

Also if you have issues with template vars see https://github.com/allenai/rslearn_projects/blob/master/esrun_data/satlas/solar_farm_oe/model.yaml for example of using the template vars.

favyen2:
I would prefer not to keep large artifacts like the annotation GeoJSON in the git repository.

PR #299, opened 2026-04-22
Update swin_pretrain to pretrain on seven modalities.

favyen2:
Also add temporal attention variant of the model.


rslp/swin_pretrain/model_temporal_attn.py
```diff
@@ -0,0 +1,206 @@
...
+        unet_out_channels: int | None = 128,
+        num_heads: int = 8,
+        mlp_ratio: float = 4.0,
+        dropout: float = 0.0,
+    ) -> None:
+        """Create new Model."""
+        super().__init__()
+        self.target_resolution_factor = target_resolution_factor
+
+        swin = torchvision.models.swin_v2_b(weights=Swin_V2_B_Weights.IMAGENET1K_V1)
```
yawenzzzz, line 114: Just curious, is there any major difference when you initialize swin with the imagenet pretrained weights vs. without?

yawenzzzz (approved):
LGTM!

rslp/swin_pretrain/model_temporal_attn.py
favyen2: Hmm I don't think I tried that for pre-training. With Satlas I think we tried it and it might've been better with ImageNet despite the pre-training data, but I'm not 100% sure. I'll see if I can quickly launch a couple experiments to check.

PR #300, opened 2026-04-27
Train improved Landsat and Sentinel-1 vessel attribute predictors

favyen2:
These use the larger training dataset that Hunter sent.

I updated the splits so that train/val have no overlapping scene IDs or vessel MMSIs.


data/landsat_vessel_attribute/README.md
```diff
@@ -1,8 +1,11 @@
-For the initial Landsat vessel attribute prediction model, several model configs are
+History:
+- 20260330: For the initial Landsat vessel attribute prediction model, several model configs are
 tested (see 20260330/). OlmoEarth-v1-Base performs the best, so it is copied to
 `config.yaml` here for use in the Landsat vessel detection pipeline. The batch size is
 reduced to 8 because 32 is too high for T4 GPU; for now it seems better to leave that
 in the config rather than adding another environment override.
+- 20260422: Hunter provided about 700K more examples. Total dataset size is about 1M now. It is a new
+  group in the same rslearn dataset path.
```
yawenzzzz, line 8: wow, this is a very big dataset, also explain why you can use more strict rule for splitting - the val % drops a lot, but for 1M it's fine

yawenzzzz (approved):
LGTM!

data/landsat_vessel_attribute/README.md
favyen2: yeah I'm surprised the performance keeps going up when we go from like 300K to 700K samples

PR #301, opened 2026-05-01
Improvements to olmoearth_evals.

favyen2:
- Remove forest_loss_driver task since anyway its performance is pretty spiky. This way we don't need lots of special cases to handle the bitemporal task.
- Move freezing config out of the per-model configs and into a separate config. We also make sure all models have just the pre-trained portion in the encoder section, so we can always target the same thing for freezing.
- Add AEF tasks. They are adapted to have 32x32 GeoTIFFs so they run faster than the original datasets we materialized (256x256).
- Move W&B best metric script from rslp.olmoearth_pretrain to rslp.olmoearth_evals.
- Remove legacy fine-tuning experiment launcher from rslp.olmoearth_pretrain.
- Add WorldCover tasks.


yawenzzzz (approved):
LGTM!

PR #303, opened 2026-05-05
Change detection data for pretraining datasets

piperwolters:
- Script to create windows from MONITRS dataset, including QA text layer
- Script to create lonlat list to be used with the create_windows_from_lonlat script 
- X Configs for MONITRS and Favyen's change detection data


data/favyen_change_detection/config.json
```diff
@@ -0,0 +1,2520 @@
+{
```
favyen2: If this is just concatenated olmoearth_pretrain modalities then I think it could go in olmoearth_pretrain instead of adding here.

favyen2 (approved):
looks good but I don't think the configs belong here, if it's a project-specific config it makes sense here but otherwise can put config with modalities concatenated in olmoearth_pretrain

PR #307, opened 2026-05-19
LFMC subsampling for olmoearth_pretrain

Farbum:
Adds scripts to create a spatially-balanced oep_eval subset of the LFMC woody and herbaceous datasets. The subsampling script (subsample_oep_eval.py) selects ~1000 train samples distributed evenly across US states, operating at the location level with a 1-year temporal cap per location to maximize geographic diversity. Val and test splits are also subsampled if they exceed 800 and 500 samples respectively (relevant for woody), otherwise all samples are included. Each selected window is tagged with "oep_eval": "" in its metadata options, following the same pattern used in the wildfire/canada_nbac dataset. A companion visualization script (visualize_oep_eval.py) produces a 2x2 map showing all original locations vs. the tagged subset for both datasets.

image

### DIFF
--- a/rslp/lfmc/subsample_oep_eval.py
+++ b/rslp/lfmc/subsample_oep_eval.py
@@ -1,4 +1,4 @@
-"""Tag a spatially-balanced subset of LFMC windows with the oep_eval tag.
+"""Tag a spatially-balanced subset of LFMC windows with a configurable tag.
 
 For train: selects ~1000 windows distributed evenly across US states, operating
 at the location level. Each selected location contributes at most 1 year of
@@ -6,7 +6,7 @@ samples (the densest 365-day window), maximizing geographic coverage.
 For val/test: tags ALL windows.
 
 Usage:
-    python subsample_oep_eval.py --dataset_path /path/to/dataset --target_train 1000 --seed 42
+    python subsample_oep_eval.py --dataset_path /path/to/dataset --tag oep_eval --target_train 1000 --seed 42
 """
 
 import argparse
@@ -220,8 +220,8 @@ def select_locations_balanced(
     return selected
 
 
-def tag_windows(windows: list[dict]) -> int:
-    """Add oep_eval tag to the given windows' metadata.json files."""
+def tag_windows(windows: list[dict], tag: str) -> int:
+    """Add the given tag to the windows' metadata.json files."""
     tagged = 0
     for i, w in enumerate(windows):
         if i % 2000 == 0 and i > 0:
@@ -229,8 +229,8 @@ def tag_windows(windows: list[dict]) -> int:
         meta_path = os.path.join(w["_dir"], "metadata.json")
         with open(meta_path) as f:
             meta = json.load(f)
-        if "oep_eval" not in meta.get("options", {}):
-            meta.setdefault("options", {})["oep_eval"] = ""
+        if tag not in meta.get("options", {}):
+            meta.setdefault("options", {})[tag] = ""
             with open(meta_path, "w") as f:
                 json.dump(meta, f)
             tagged += 1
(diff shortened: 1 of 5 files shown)


favyen2 (approved):

rslp/lfmc/subsample_oep_eval.py
```diff
@@ -0,0 +1,474 @@
...
+        with open(meta_path) as f:
+            meta = json.load(f)
+        if "oep_eval" not in meta.get("options", {}):
+            meta.setdefault("options", {})["oep_eval"] = ""
+            with open(meta_path, "w") as f:
+                json.dump(meta, f)
+            tagged += 1
+        else:
+            tagged += 1
+    return tagged
```
APatrickJ, line 241: Since we always tag, doesn't this always just return `len(windows)`?

rslp/lfmc/subsample_oep_eval.py
```diff
@@ -0,0 +1,474 @@
+"""Tag a spatially-balanced subset of LFMC windows with the oep_eval tag.
+
+For train: selects ~1000 windows distributed evenly across US states, operating
+at the location level. Each selected location contributes at most 1 year of
+samples (the densest 365-day window), maximizing geographic coverage.
+For val/test: tags ALL windows.
```
APatrickJ, line 10: Looks like we subsample val if > 800 and test if > 500, should we update this?

rslp/lfmc/subsample_oep_eval.py
```diff
@@ -0,0 +1,474 @@
...
+        }
+
+    if args.dry_run:
+        print("\n[DRY RUN] No tagging performed.")
+        return
+
+    # Tag windows
+    print("\nTagging selected train samples...")
+    train_tagged = tag_windows(selected_train)
+    print(f"  Tagged {train_tagged} train samples")
```
APatrickJ, line 439: Running this twice with different seeds/targets will leave the union of both runs tagged. We could take the same approach as `create_oep_eval.py` for the wildfire use case and handle this by reading the prior `oep_eval_manifest.json` and clearing old tags before re-tagging.

Or we could document that this is a run-once script.

rslp/lfmc/subsample_oep_eval.py
Farbum: yes good catch, created the summary before hardcoding limits

rslp/lfmc/subsample_oep_eval.py
Farbum: Good catch, fixed

rslp/lfmc/subsample_oep_eval.py
```diff
@@ -0,0 +1,474 @@
...
+    if unsettled:
+        per_state_quota = remaining_target / len(unsettled) if unsettled else 0
+        for state in unsettled:
+            locs = state_to_locs[state]
+            locs_with_count = [(lk, loc_one_year_count[lk]) for lk in locs]
+
+            # Shuffle for randomness, then sort by count (prefer dense locations)
+            shuffled = list(locs_with_count)
+            rng.shuffle(shuffled)
+            shuffled.sort(key=lambda x: x[1], reverse=True)
```
APatrickJ, line 199: Why not just shuffle here? The shuffle-then-sort is confusing since the stable sort immediately overrides the shuffle order, making the shuffle only a tiebreaker for same-count locations.

APatrickJ (approved):

rslp/lfmc/subsample_oep_eval.py
Farbum: It is to break ties randomly, nothing crucial to the subsampling though

PR #308, opened 2026-05-20
Detect land cover changes in DRC from Feb to May 2026

favyen2:


one_off_projects/2026_05_19_drc_worldcover/README.md
```diff
@@ -0,0 +1,153 @@
...
+The model outputs 12-band float32 softmax probabilities (`output_probs: true`),
+one band per worldcover class.
+
+### 5. Detect land cover changes
+
+Compare the pre-period (dec2025, jan2026, feb2026) and post-period (may2026)
+probability outputs to find locations where tree/forest cover was confidently
+lost. For each pixel the **minimum** tree probability across all pre-period
+months must exceed `--pre_threshold` (0.75), and the May probability must fall
+below `--post_threshold` (0.25). The pre-period confident mask is eroded by
```
piperwolters, line 116: Does this erode_pixels arg just ignore pixels at edges between forest and other classes?

one_off_projects/2026_05_19_drc_worldcover/worldcover_4ts.yaml
```diff
@@ -0,0 +1,341 @@
...
+              metric:
+                class_path: torchmetrics.classification.MulticlassRecall
+                init_args:
+                  num_classes: 12
+                  average: null
+              class_idx: 11
+    batch_size: 16
+    num_workers: 32
+    default_config:
+      crop_size: 16
```
piperwolters, line 275: Isn't crop size usually 64, what is reasoning to make it 16 here?

one_off_projects/2026_05_19_drc_worldcover/README.md
```diff
@@ -0,0 +1,153 @@
...
+- `timestamp` — approximate date of the change (default `2026-04-01`)
+- `num_pixels` — number of pixels in the connected component
+
+Override pre-period groups with `--pre_groups dec2025,jan2026,feb2026`.
+
+To detect loss of a different class, pass `--pre_class_index` (default 8 =
+tree). Class indices: 0=bare, 1=burnt, 2=crops, 3=fallow, 4=grassland,
+5=lichen\_moss, 6=shrub, 7=snow\_ice, 8=tree, 9=urban, 10=water, 11=wetland.
+
+Note: it may be useful to remove forest loss events after the initial GeoJSON is
```
piperwolters: What do you mean by remove the forest loss events?

one_off_projects/2026_05_19_drc_worldcover/detect_changes.py
```diff
@@ -0,0 +1,299 @@
...
+    if pre_groups is None:
+        pre_groups = ["dec2025", "jan2026", "feb2026"]
+
+    dataset = Dataset(UPath(ds_path))
+    band_names = dataset.layers["output"].band_sets[0].bands
+
+    # Load pre-period windows keyed by spatial location, one list per key.
+    pre_by_key: dict[str, list[Window]] = {}
+    for group in pre_groups:
+        windows = dataset.load_windows(groups=[group], workers=128, show_progress=True)
```
piperwolters: Do you want this to use the `workers` arg that is passed in here rather than hardcoding?

piperwolters (approved):

one_off_projects/2026_05_19_drc_worldcover/README.md
favyen2: clarified that it will apply that many iterations of binary erosion on the change mask.

so the initial change mask will just be a per-pixel comparison, and edge artifacts will be removed by applying binary erosion.

one_off_projects/2026_05_19_drc_worldcover/worldcover_4ts.yaml
favyen2: I updated the README to document this, I used patch size 2 I think because I wanted the model to perform better with small-scale changes. And then must've set crop size to 16 because 64x64 is too big for patch size 2 to be efficient.

one_off_projects/2026_05_19_drc_worldcover/detect_changes.py
favyen2: Fixed, thanks, I think I changed this to make the window loading faster since I wanted to use more workers for window loading than detecting changes due to memory limitation in the latter part, but I reverted it to just use the workers argument, I think it could make sense to add --load_workers vs --change_workers (two arguments) but probably unnecessary for this one-off project.

one_off_projects/2026_05_19_drc_worldcover/README.md
favyen2: Hmm shoot I'm not sure, I think it might be about some transitions having a high rate of false positives like tree->shrubs. I updated the note to say that there might've been certain transitions like tree->shrub that were false positive and were manually removed but I also included in the README note that I don't remember exactly.

favyen2:
I cleaned up the re-projection code too, so it will just do transformation to absolute pixel coordinates once instead of doing that but also doing the reverse (to get the pre/post class). And added comments.

PR #309, opened 2026-05-21
rslp.olmoearth_evals: more eval tasks and use in memory dataset

favyen2:
- Add tasks
  - forest_loss_driver (added back, but just do one forward pass instead of special case)
  - hls_burn_scars
  - kenya_intercropping
  - surface_fuels
- Add new versions of tasks that have fewer samples so evaluation is fast:
  - small_eurosat
  - small_mangrove_ts
  - small_mangrove_mm
  - small_sentinel2_vessel_length
  - small_sentinel2_vessel_type
- Other changes:
  - Use in memory dataset for tasks that are small enough, to substantially speed up training
  - Clean up the decoders in `olmoearth.py` to make them more consistent and raise error if invalid decoder is specified


piperwolters:
@claude review

favyen2:
@claude I don't see this "val and test are copied in full" can you clarify about it?

favyen2:
@claude okay can you take another look now? I fixed a few things

piperwolters (approved):
Eval tasks and improvements allow for us to benchmark various finetuning strategies in rslearn, which will ultimately benefit partner tasks.

Claude reviewed, some cleanups were pushed. Looks good to me!

PR #314, opened 2026-06-02
Update Landsat classification model

yawenzzzz:
1. add annotation tool, focusing on labelling hard negatives
2. add Landsat v1.0.0 (new model trained with old annotations + 2K new annotation, use OlmoEarth v1.2)
3. add feedback workflow, when we have a model in integration, wait for initial results (ideally 15 days or even longer), add feedback samples back to training, to help further remove FPs


data/landsat_vessels/config_classifier_20260616.yaml
```diff
@@ -0,0 +1,103 @@
+model:
```
favyen2, line 1: I think you can just leave the config you decided to use here and the rest can either be deleted (still will be retained in git history) or move to subfolder like `data/landsat_vessels/20260908_exploration/`.

rslp/landsat_vessels/annotations/README.md
```diff
@@ -0,0 +1,422 @@
+# Landsat vessel classifier — annotation rounds
+
+Plans and records for re-annotating the vessel classifier dataset. Each round gets a
+section here; scripts live in `rslp/landsat_vessels/scripts/`.
+
```
favyen2, line 5: maybe there can be a short blurb here saying that there were previous annotations too and we train on those but afterward new annotations would be from this new app.

rslp/landsat_vessels/evaluation/smoke_test.py
```diff
@@ -0,0 +1,79 @@
+"""Smoke test: run the predict pipeline on T1/T2 scenario-check scenes."""
```
favyen2, line 1: Is this just to test to verify it works? If so I don't think it needs to be in the repository -- can remove or convert it into an actual unit test.

rslp/landsat_vessels/scripts/verify_round1_windows.py
```diff
@@ -0,0 +1,130 @@
+"""Verify the round-1 windows are trainable, by reading them the way training will.
```
favyen2, line 1: Are all of these scripts useful? If so can keep them but need a README.md to document them I think, since there are so many.

rslp/landsat_vessels/README.md
```diff
@@ -17,3 +17,76 @@ python -m rslp.main common beaker_train --image_name favyen/rslphelios2 --config
 ```
 python -m rslp.main olmoearth_pretrain launch_finetune --olmoearth_checkpoint_path /weka/dfive-default/helios/checkpoints/favyen/v0.2_base_latent_mim_128_alldata_random_fixed_modality_0.5/step320000 --patch_size 4 --encoder_embedding_size 768 --image_name favyen/rslphelios3 --config_paths+=data/helios/v2_landsat_vessels/finetune_classifier.yaml --cluster+=ai2/ceres-cirrascale --project_name 2025_06_26_helios_finetuning --run_name v2_landsat_vessel_classification_helios_base_ps4_add_prob_threshold
 ```
+
+## Classifier training results (2026-09-08)
```
favyen2, line 21: I think this README will need some summary talking about what the sub-modules are for like annotation coming from existing system + new annotations using rslp.landsat_vessels.annotations, which uses rslp.landsat_vessels.feedback to initialize it, and rslp.landsat_vessels.evaluation helps with X.

rslp/landsat_vessels/config.py
```diff
@@ -1,6 +1,41 @@
 """This module contains the configuration for the Landsat Vessel Detection pipeline."""
 
 import json
+import os
+from pathlib import Path
+
+# Config paths are anchored to the directory that actually holds the `data/` tree rather
```
favyen2, line 7: Can we just assume the user is running from repo root here instead of needing to detect the repo root automatically?

rslp/landsat_vessels/transforms.py
```diff
@@ -0,0 +1,78 @@
+"""Transforms for the Landsat vessel models."""
```
favyen2, line 1: Can we use the existing Pad transform or it doesn't work?

PR #315, opened 2026-06-03
Lazy all-crops dataset for memory-safe Sentinel-1 prediction

rpw1134:
## Problem
The `load_all_crops` prediction path reads the **entire window** into memory and then slices crops out of it. For a full Sentinel-1 scene the window spans the whole ~25000×16700 px image (6 channels float32, ~10GB raw / ~28GB after `Concatenate`+`Normalize` copies), which OOM-kills the dataloader worker on the first batch — before any detection is produced.

## Fix
Read each crop's bounds from disk on demand instead of materializing the full window.

- **`rslp/utils/lazy_all_crops.py`** (new):
  - `LazyAllCropsDataset` — map-style (mirrors `InMemoryAllCropsDataset`) but does not cache the window; each `__getitem__` reads only one crop bbox via rslearn's `read_data_input`. No full-window array, no in-memory slicing.
  - `LazyAllCropsDataModule` — `RslearnDataModule` subclass; `setup()` swaps `IterableAllCropsDataset` for `LazyAllCropsDataset`.
- **`data/sentinel1_vessels/config.yaml`** — point `data.class_path` at `LazyAllCropsDataModule`.

`rslearn` core is unchanged. Crops/model/outputs are identical — only the memory profile differs.

## Scope
- Only affects the Sentinel-1 vessel pipeline (its config is the only one referencing the new module). Sentinel-2 / Landsat unchanged.
- Training: the `train` split doesn't use `load_all_crops` (no change); `val`/`test` produce identical results with lower memory.

## Verification
Ran the failing scene + 2 historicals through the container on an n1-standard-8 + T4 (same machine type as the prod node):
- Before: predict-stage memory climbs to ~29GB → kernel OOM-kill of `pt_data_worker`.
- After: `LazyAllCropsDataset (3402 crops, on-demand reads)`, predict-stage memory steady ~3–4GB, no OOM. Full pipeline returns HTTP 200 with 17 detections (detector → attribute model).

### DIFF
--- a/data/sentinel1_vessels/config.yaml
+++ b/data/sentinel1_vessels/config.yaml
@@ -38,7 +38,7 @@ model:
       remap_prefixes:
         - ["backbone.backbone.backbone.", "encoder.0.encoder.model."]
 data:
-  class_path: rslearn.train.data_module.RslearnDataModule
+  class_path: rslp.utils.lazy_all_crops.LazyAllCropsDataModule
   init_args:
     path: /weka/dfive-default/rslearn-eai/datasets/sentinel1_vessels/dataset_v1/20250521/
     inputs:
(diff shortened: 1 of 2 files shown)


favyen2:
closing since it is added to rslearn instead

PR #317, opened 2026-06-08
Ryan/sentinel1 fix crop fnames

rpw1134:
Frames were being orphaned in the Sentinel1 pipeline, causing no images to render.

Added `detection.crop_fnames = crop_fnames` after the per-band loop, so the written file paths are actually stored on the detection object


favyen2 (approved):
looks good and I will take a look at the test error later today

PR #321, opened 2026-06-09
Hadriens/mapbiomas

Farbum:
Adding mapbiomas subsampling for olmoearth_pretrain evals

Subsample results:
image

Rslearn model results:
image


rslp/lfmc/compute_target_stats.py
```diff
@@ -1,285 +0,0 @@
-"""Compute mean/std of LFMC target values over a split (and optional tag).
```
favyen2, line 1: Are these files supposed to be deleted here?

favyen2: I think something happened because now those LFMC configs were merged into master but the git history here includes adding them and then deleting them so now they are being deleted.

Farbum: Yes these LFMC files were part of another branch (LFMC one) that was merged. They were wrongly added to this PR so I removed them

data/helios/mapbiomas/sparse_config/model.yaml
```diff
@@ -33,77 +30,89 @@ model:
...
+          segment:
+            class_path: rslearn.train.tasks.segmentation.SegmentationTask
             init_args:
-              nodata_value: -1
-              metrics: ["rmse", "r2"]
-              #scale_factor: 0.0024038
-              target_mean: 108.389520
-              target_std: 36.616698
-
+              num_classes: 12
```
favyen2, line 55: Why are there only 12 classes here? in the dataset if there are other class IDs will they cause CUDA error?

one_off_projects/2026_06_01_mapbiomas/subsampling/__init__.py
```diff
@@ -0,0 +1 @@
+"""MapBiomas balanced subsampling utilities."""
```
favyen2, line 1: Add README in rslp/mapbiomas/ about where the task was downloaded from and basic overview of the data.

rslp/lfmc/compute_target_stats.py
favyen2: Yeah but now after merging in master, this PR is deleting those files from the other PR. Because of the way the git history is merged across the two branches.

data/helios/mapbiomas/sparse_config/model.yaml
Farbum: Yes sparse setting has 12 classes. added a readme for info

rslp/lfmc/compute_target_stats.py
Farbum: should be good now

favyen2 (approved):

PR #323, opened 2026-06-22
Sentinel-1 vessels: tile scenes into detector windows to cut materialize memory

rpw1134:
## Motivation

The S1 vessels sidecar OOMs intermittently on large S1C/S1D scenes. The detector runs on a single full-scene `Window`, so `materialize_dataset` builds a full-scene composite (FirstValidCompositor → whole-scene read); measured anon peak ~27 GiB, above an n1-standard-8 node's ~25.9 GiB allocatable → node OOM.

## Change

- Split each scene into a grid of disjoint `DETECTOR_TILE_SIZE` (2048px) detector windows; each tile materializes and runs the detector independently, so peak memory scales with the tile, not the scene.
- Add cross-tile suppression (`_merge_tile_detections`): a vessel near a tile boundary can be detected in two tiles, so dedup by distance NMS per scene. Reuses `distance_nms`, factored out of `NMSDistanceMerger._apply_nms` (behavior unchanged; `merge()` still delegates to it).

Draft: detection-equivalence validation against the full-scene path is pending. See inline notes.

### DIFF
--- a/rslp/sentinel1_vessels/predict_pipeline.py
+++ b/rslp/sentinel1_vessels/predict_pipeline.py
@@ -490,20 +490,33 @@ def get_vessel_detections(
     )
     with time_operation(TimerOperations.MaterializeDataset):
         materialize_dataset(ds_path, materialize_pipeline_args)
-    # Verify the target image and both historical images were materialized for each
-    # window before running prediction.
-    for window in windows:
-        if not window.is_layer_completed(SENTINEL1_LAYER_NAME):
+
+    # Keep only tiles that have the target and both historical images materialized. A
+    # scene's bounds form a rectangle around its slanted SAR swath, so tiles over the
+    # nodata margin (or outside the historicals' overlap) never materialize; drop those
+    # tiles, deleting their window so the detector does not read missing layers.
+    complete_windows: list[Window] = []
+    complete_scene_idxs: list[int] = []
+    for window, scene_idx in zip(windows, window_scene_idxs):
+        has_all_layers = window.is_layer_completed(SENTINEL1_LAYER_NAME) and all(
+            window.is_layer_completed(HISTORICAL_LAYER_NAME, group_idx)
+            for group_idx in range(HISTORICAL_GROUP_COUNT)
+        )
+        if has_all_layers:
+            complete_windows.append(window)
+            complete_scene_idxs.append(scene_idx)
+        else:
+            window.window_root.fs.rm(window.window_root.path, recursive=True)
+    windows = complete_windows
+    window_scene_idxs = complete_scene_idxs
+
+    materialized_scene_idxs = set(window_scene_idxs)
+    for scene_idx in range(len(scene_datas)):
+        if scene_idx not in materialized_scene_idxs:
             raise ValueError(
-                f"window {window.name} does not have Sentinel-1 layer completed"
+                f"scene {scene_idx} has no fully-materialized tiles; the detector requires "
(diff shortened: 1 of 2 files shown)


rslp/sentinel1_vessels/predict_pipeline.py
```diff
@@ -66,6 +68,15 @@
 # Band order as configured in dataset.
 BAND_NAMES = ["vv", "vh"]
 
+# Side length in pixels of the tiles each scene is split into for detection. Each tile is
+# materialized and run through the detector independently, so peak memory scales with the
+# tile size rather than the full scene.
+DETECTOR_TILE_SIZE = 2048
```
rpw1134, line 74: Per your note — tiles are disjoint here (no overlap). At 2048² a vessel landing on a seam is uncommon, so I skipped overlap/halo entirely and rely on the cross-tile NMS below to catch the few that do split.

rslp/sentinel1_vessels/predict_pipeline.py
```diff
@@ -467,7 +543,7 @@ def get_vessel_detections(
 
             detections.append(detection)
 
-    return detections
+    return _merge_tile_detections(detections)
```
rpw1134, line 559: This is the across-window NMS you flagged as still needed even without overlap: a vessel straddling a tile boundary can be detected in both tiles, so we dedup per scene here. Centers are in the shared scene projection (all tiles use the same `Projection`), so they're directly comparable across tiles — no offset bookkeeping.

rslp/utils/nms.py
```diff
@@ -15,6 +15,69 @@
 DEFAULT_DISTANCE_THRESHOLD = 10
 
 
+def distance_nms(
```
rpw1134, line 18: On reusing `NMSDistanceMerger`: the class is a per-window `CropPredictionMerger` hook (`merge(window, outputs, layer_config)`), so it's only ever invoked within a single window and can't dedup across windows. The NMS *algorithm* is coordinate-agnostic, though, so I factored it into this `distance_nms`; `_apply_nms` now delegates to it (behavior unchanged) and the cross-tile pass calls it directly. So: reused the algorithm, no new merger class.

rpw1134:
## Benchmark (ryan-dev, T4; high-peak S1D `…75D5`; rslearn pinned to deployed #667, so this isolates the tiling effect)

| Metric | Tiled (this PR) | Full-scene baseline |
|---|---|---|
| Materialize peak | **13.3 GiB** (anon 10.2) | ~20.4 GiB |
| Wall-clock | 27 min | ~30 min |
| Box stability | clean | wedges a 30 GB box (swap) |
| Detections | 111 | 114 |

**Memory goal met** — peak well under an n1-standard-8's ~25.9 GiB, runs end-to-end without the OOM. Tiling confirmed (scene → ~180 tiles); empty-tile drop + cross-tile NMS execute.

**⚠️ Open / blocking before ready:** detection equivalence is **106/114 matched within 50 m** (8 missing, 5 extra) — not yet equivalent. Need to localize the ~13 discrepant detections:
- *missing* likely from dropped historical-edge tiles and/or seam cuts → fix: materialize a nodata-filled historical for partial tiles instead of dropping them;
- *extra* likely seam-split duplicates the cross-tile NMS didn't merge → revisit seam handling / NMS threshold.

Memory peak is also driven by `workers=32` materializing 32 tiles at once — lowering that pushes the peak toward the single-tile floor.

favyen2:
I think we addressed the worst memory peak by separating the band set so each band is stored separately, the tiling would still help to make the memory usage really low but I think it introduced some issues.

PR #326, opened 2026-06-30
Update Sentinel-1 vessel detection pipeline to use the 1hist model

favyen2:
So this way it just uses one historical Sentinel-1 image.

PR #333, opened 2026-07-15
Sentinel-1: per-request score threshold with env-configurable default

rpw1134:
Makes the Sentinel-1 detector's `score_threshold` a per-request override with an env-configurable default, resolved into a `run_model_predict` decode override.

> [!NOTE]
> **Stacked on #334** (drop un-croppable edge detections). Merge #334 first.

**Operating model (two default sources, by execution context)**
- **Outside the predict pipeline** (direct model run / eval): `config.yaml` `score_threshold` (**0.7**) applies. The pipeline always overrides it, so it never affects API traffic.
- **Inside** (the API/sidecar): `api_main.SENTINEL1_SCORE_THRESHOLD` — from the `SENTINEL1_SCORE_THRESHOLD` env var, **0.7** fallback — is the default; a request's `score_threshold` overrides it.

Resolution: `request.score_threshold  ??  SENTINEL1_SCORE_THRESHOLD (env, else 0.7)` → injected as `--data.init_args.task.init_args.tasks.detect.init_args.score_threshold`, so a request can raise **or lower** the cutoff (including below the config value) with no rebuild.

**Changes**
- Rename `confidence_threshold` → **`score_threshold`** across the request field, `predict_pipeline`, and `get_vessel_detections`.
- `predict_pipeline`/`get_vessel_detections` take a **required** `score_threshold` and always inject the decode override; removed the post-detection filter.
- `api_main`: `SENTINEL1_SCORE_THRESHOLD` module default (env, 0.7 fallback); request wins via an explicit `is not None` check (so `score_threshold=0.0` is honored, not swallowed).
- `Dockerfile`: no baked default; set `SENTINEL1_SCORE_THRESHOLD` at deploy time (kubectl/terraform) to change the API default without a rebuild.
- Tests: `test_api_score_threshold.py` (default / configurable / request-overrides / 0.0-honored); updated the three sentinel1 `predict_pipeline` call sites for the required arg.

**Operating summary:** do-nothing → 0.7; change the default → env on the sidecar, no rebuild, no sat change; one-off → request `score_threshold`.

**Other PRs:** #334 (edge fix) and #326 (monorepo tag bump) need no changes — the monorepo sends no threshold.


favyen2 (approved):
looks good to me, if you want to lower it further and control filtering downstream that sounds fine too

hunterp (approved):

rslp/sentinel1_vessels/api_main.py
```diff
@@ -197,6 +200,10 @@ async def get_detections(
                 tasks=[task],
                 scratch_path=scratch_path,
             )[0]
+        if info.confidence_threshold is not None:
```
hunterp: this should probably be in a different place in the pipeline; since we 
1. don't want to run the attribute model on detections that will be filtered out
2. don't want to write image chips to disk for detections that will be filtered out.

rslp/sentinel1_vessels/api_main.py
```diff
@@ -97,13 +97,16 @@ class Sentinel1Request(BaseModel):
             same orbit direction as the target image.
         crop_path: Optional; Path to save the cropped images.
         scratch_path: Optional; Scratch path to save the rslearn dataset.
+        confidence_threshold: Optional; drop detections scoring below this value.
+            If unset, all decoded detections above the model's 0.5 floor are returned.
```
hunterp: we should support configuring this via an env var so you're not passing it in every request.   The env var should be the default unless it's overriden in a particular request. 

The idea here is that: 
1. We configure the standard confidence threshold at an enviornment level, this is configured in the deployment and is the default level that all runs operate at.  nothing more is needed beyond the deployment.   If no value is provided by the enviornment, a reasonable backup value is used (e.g. 0.7)
2. You can overrride that env var on a per-request basis; so if you do want to run a scene at a lower threshold (e.g. 0.5) you can do that without re-deploying.

hunterp (dismissed):
this should support providing the value in either 
1. the request parameter
2. an environment variable 

Where (1) would override (2)

rslp/sentinel1_vessels/api_main.py
```diff
@@ -30,6 +30,14 @@
 SENTINEL1_HOST = os.getenv("SENTINEL1_HOST", "0.0.0.0")
 SENTINEL1_PORT = int(os.getenv("SENTINEL1_PORT", 5555))
 
+# Default detection score cutoff, applied unless a request supplies its own
+# confidence_threshold. The model decodes down to its 0.5 floor; this drops detections
+# below the cutoff. Unset => no default filtering (all decoded detections returned).
+_env_score_threshold = os.getenv("SENTINEL1_SCORE_THRESHOLD")
+DEFAULT_CONFIDENCE_THRESHOLD = (
```
favyen2: I don't quite like how this isn't actually a constant. Could we just remove all the global variables and resolve everything below in `get_detections`? I don't think reading the env var on each request is going to impact performance at all.

rslp/sentinel1_vessels/Dockerfile
```diff
@@ -4,6 +4,8 @@ FROM base-image:latest
 # Environment variables
 ENV PYTHONPATH="/opt/rslearn_projects:${PYTHONPATH}"
 ENV SENTINEL1_PORT=5555
+# Default detection score cutoff; a request's confidence_threshold overrides it.
+ENV SENTINEL1_SCORE_THRESHOLD=0.7
```
favyen2: Adding this here results in two defaults -- one in `config.yaml` and one in `Dockerfile` Can we just have the one in `config.yaml` and remove the env var from here? AFAIK the code will already fallback to the config file threshold if the env var is not set. If downstream users want to set the env var I think they can set it directly when running the Docker container.

I think it'd be fine to have a comment at the top of the Dockerfile saying there's another env var `SENTINEL1_SCORE_THRESHOLD` that can be set to configure default score threshold for requests.

rslp/sentinel1_vessels/predict_pipeline.py
```diff
@@ -591,6 +591,7 @@ def run_attribute_model(
 def predict_pipeline(
     tasks: list[PredictionTask],
     scratch_path: str | None = None,
+    confidence_threshold: float | None = None,
```
favyen2: If the override approach I suggested on slack works then I think that'd be preferable so that the user can call it with confidence threshold lower than the one in the model config file. Currently if you set threshold for one request to 0.1 it would still cut off at 0.5.

Override would be adding the threshold to `run_model_predict` via the extra arguments. Something like `--data.init_args.task.init_args.tasks.detect.init_args.score_threshold=0.5`.

rslp/sentinel1_vessels/Dockerfile
hunterp: Agree, we should not set this value here.   The env var can be set by the deployment environment (e.g. k8s in the sat-service environment)

rslp/sentinel1_vessels/predict_pipeline.py
hunterp: also nit pick:  This won't be `None` here; since  you either get the value from the request OR the env var.  It's never not set.

tests/unit/sentinel1_vessels/__init__.py
hunterp, line 1: 👏 you 👏 don't 👏 need 👏  empty 👏  init.py 👏  files 👏 

You need test files that have unique names.

rslp/sentinel1_vessels/predict_pipeline.py
rpw1134: Responding to your comment above and this one: 

Hunter and I had an idea of a three tiered threshold hierarchy, which is what these changes are based on:
1. The model always runs with the the threshold in yaml
2. We return detections above the threshold defined in a request (if it exists), and if not then the rslearn_projects Dockerfile env var. This lets us exec into the pod and change the confidence threshold whenever but still gives clients power

What I've gathered from your message is you're suggesting we cut out the Dockerfile env var and run the pipeline with the yaml threshold. So a user can exec into the pod, change the yaml variable, and get similar results. Clients can send a request variable to directly change what the pipeline runs. Is this accurate?

If so, I agree. One less variable, a single source of truth (confidence runs in pipeline rather than filtering after), same results, and allows clients to request even lower thresholds. @hunterp , is this good with you?

rslp/sentinel1_vessels/api_main.py
```diff
@@ -192,10 +196,20 @@ async def get_detections(
 
     try:
         logger.info("Processing request with input data.")
+        # Effective cutoff: request value, else the SENTINEL1_SCORE_THRESHOLD env default,
+        # else none (return all decoded detections).
+        env_threshold = os.getenv("SENTINEL1_SCORE_THRESHOLD")
```
hunterp: this should not be inline.  at minimum this should be wherever other env vars are loaded from.   probably in a config.py or settings.py kind of file.  

Also this should provide a default value.

favyen2 (approved):
thanks for iterating on this, it looks great to me!

PR #334, opened 2026-07-15
Sentinel-1: drop un-croppable edge detections instead of crashing the scene

rpw1134:
`run_attribute_model` raised `ValueError` if *any* detection's crop window couldn't be materialized, failing the **entire scene**. That happens for a detection close enough to the scene edge that its 128px crop runs off the imagery — there's neither an attribute prediction nor a crop image to produce for it (both are read from the same crop raster).

**Fix:** drop those detections (and delete their windows so the attribute model skips them) instead of raising, keeping `detections` and `windows` index-aligned.

**Why now:** this raise has existed since the pipeline's first commit. It only surfaces at lower detection score thresholds — a higher threshold admits few detections, so none land at the edge. Lowering the floor (see the stacked confidence-threshold PR) admits more, so one reliably hits the edge. The bug was always there; the threshold just exposes it.

Standalone this is a no-op at today's 0.9 threshold; it's exercised by the confidence-threshold PR stacked on top, whose integration test runs at the 0.5 floor.

### DIFF
--- a/rslp/sentinel1_vessels/predict_pipeline.py
+++ b/rslp/sentinel1_vessels/predict_pipeline.py
@@ -538,12 +538,29 @@ def run_attribute_model(
     )
     materialize_dataset(ds_path, materialize_pipeline_args)
 
-    # Verify that no window is unmaterialized.
-    for window in windows:
-        if not window.is_layer_completed(SENTINEL1_LAYER_NAME):
-            raise ValueError(
-                f"window {window.name} does not have materialized Sentinel-1 image"
-            )
+    # Drop detections whose crop window could not be materialized -- e.g. a detection so
+    # close to the scene edge that its crop runs off the imagery. Without a materialized
+    # crop there is neither an attribute prediction nor a crop image to produce, so the
+    # detection cannot be reported. Delete the unusable windows so the attribute model
+    # does not run on them, and keep detections/windows index-aligned.
+    kept_detections: list[VesselDetection] = []
+    kept_windows: list[Window] = []
+    for detection, window in zip(detections, windows):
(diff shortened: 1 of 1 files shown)


rslp/sentinel1_vessels/predict_pipeline.py
```diff
@@ -538,12 +538,29 @@ def run_attribute_model(
...
+    # detection cannot be reported. Delete the unusable windows so the attribute model
+    # does not run on them, and keep detections/windows index-aligned.
+    kept_detections: list[VesselDetection] = []
+    kept_windows: list[Window] = []
+    for detection, window in zip(detections, windows):
+        if window.is_layer_completed(SENTINEL1_LAYER_NAME):
+            kept_detections.append(detection)
+            kept_windows.append(window)
+        else:
+            window.window_root.fs.rm(window.window_root.path, recursive=True)
```
favyen2: Could we just use `len(detections) == 0` check below so we don't need to clear out those windows? I think I don't like the `rm` because it's just not something we usually do elsewhere. And I think the main thing is we need to return if there are no windows with data since otherwise Lightning will raise an error about empty data loader, but I think the attribute model will ignore the windows without the completed layer already.

favyen2 (approved):
Makes sense to me, thanks for the fix. I've run into this occasionally before but I don't think I really understand why it comes up -- even if the detection is on the edge of the image, shouldn't it still be able to materialize that layer, since the dataset configuration for the attribute model has SpaceMode.MOSAIC (default) and not SpaceMode.CONTAINS? Or is it somehow detecting vessels at locations covered by historical images but not by the current image? Would be nice to understand this if there's time to probe a bit but I think it's also fine to merge.

PR #335, opened 2026-07-23
Infrastructure to compute embeddings at scale on Beaker

favyen2:


rslp/large_scale_embeddings/predict_pipeline.py
```diff
@@ -0,0 +1,620 @@
...
+            ds_path, materialize_pipeline_args=MATERIALIZE_PIPELINE_ARGS
+        )
+
+        # Run the model only if at least one window has materialized imagery.
+        completed_fnames = list(
+            ds_path.glob(
+                f"windows/{PREDICTION_GROUP}/*/layers/{SENTINEL2_LAYER}/completed"
+            )
+        )
+        if len(completed_fnames) == 0:
```
APatrickJ: I think we could replace this with:

```suggestion
        if not any(window.is_layer_completed(SENTINEL2_LAYER) for window in windows):
```

rslp/large_scale_embeddings/predict_pipeline.py
```diff
@@ -0,0 +1,620 @@
...
+    "B04",
+    "B05",
+    "B06",
+    "B07",
+    "B08",
+    "B8A",
+    "B09",
+    "B11",
+    "B12",
+]
```
APatrickJ: Optionally, we could get these from the dataset because it looks like they need to be identical to the config anyway. Something like this once we construct the dataset:

```python
dataset.layers[SENTINEL2_LAYER].band_sets[0].bands
```

rslp/large_scale_embeddings/predict_pipeline.py
```diff
@@ -0,0 +1,620 @@
...
+
+OUTPUT_LAYER = "output"
+EMBEDDING_DIM = 128
+# The dataset configs use num_bands, for which rslearn generates band names B0, B1,
+# etc.
+OUTPUT_BANDS = [f"B{band_idx}" for band_idx in range(EMBEDDING_DIM)]
```
APatrickJ: Similarly,

```python
dataset.layers[OUTPUT_LAYER].band_sets[0].bands
```

data/large_scale_embeddings/s2_s1.yaml
```diff
@@ -0,0 +1,114 @@
...
+              # (--checkpoint_path).
+              checkpoint_path: /path/to/checkpoint_dir
+              # patch_size 1 yields one 128-dimensional embedding per 10 m pixel.
+              # Overridden by the prediction pipeline (--patch_size).
+              patch_size: 1
+              embedding_size: 128
+              # Output the 128-dim spatial register bottleneck latents (one per
+              # patch in dynamic-grid mode) rather than the 768-dim encoder patch
+              # tokens.
+              use_register_bottleneck_output: true
```
APatrickJ, line 22: Since this is fairly new I think we should make sure we have it in an rslearn release and also update the rslearn dependency in `requirements.txt` to set that version as the lower bound

rslp/large_scale_embeddings/write_jobs.py
```diff
@@ -0,0 +1,318 @@
...
+                    .shp
+                )
+            if len(zone_geojson_shapes) == 0:
+                continue
+
+        user_bounds_in_proj: PixelBounds | None = None
+        if wgs84_bounds is not None:
+            dst_geom = STGeometry(
+                WGS84_PROJECTION, shapely.box(*wgs84_bounds), None
+            ).to_projection(projection)
```
APatrickJ: I think this may break for certain AOIs, e.g. `(36.0, -2.0, 38.0, 1.0)`

Possible fix could be to add a small helper like this:

```python
def bounds_near_zone(wgs84_bounds, utm_zone) -> bool:
    min_lon, _, max_lon, _ = get_wgs84_bounds(utm_zone)
    return (
        wgs84_bounds[2] >= min_lon - ZONE_LONGITUDE_MARGIN
        and wgs84_bounds[0] <= max_lon + ZONE_LONGITUDE_MARGIN
    )
```

note: `s/GEOJSON_ ZONE_LONGITUDE_MARGIN/ZONE_LONGITUDE_MARGIN/g`

and something like this here:

```suggestion
            if not bounds_near_zone(wgs84_bounds, utm_zone):
                continue
            dst_geom = STGeometry(
                WGS84_PROJECTION, shapely.box(*wgs84_bounds), None
            ).to_projection(projection)
```

and the above code could simplify to:

```python
            zone_geojson_shapes = []
            for shp in geojson_shapes:
                if not bounds_near_zone(shp.bounds, utm_zone):
                    continue
                zone_geojson_shapes.append(
                    STGeometry(WGS84_PROJECTION, shp, None)
                    .to_projection(projection)
                    .shp
                )
```

requirements.txt
```diff
@@ -1,3 +1,4 @@
+global-land-mask>=1.0.0
```
APatrickJ: Should this go in requirements-extra.txt?

rslp/large_scale_embeddings/tiling.py
```diff
@@ -0,0 +1,162 @@
...
+        bounds: the pixel bounds to divide into crops. Each value must be a multiple
+            of crop_size.
+        crop_size: the size of each crop in pixels.
+        wedge: the zone wedge from get_zone_wedge; computed if not provided.
+
+    Returns:
+        list of pixel bounds of the crops to process.
+    """
+    for value in bounds:
+        assert value % crop_size == 0
```
APatrickJ: We likely want to raise a `ValueError` here

rslp/large_scale_embeddings/tiling.py
```diff
@@ -0,0 +1,162 @@
...
+    if wedge is None:
+        wedge = get_zone_wedge(projection.crs, projection.x_resolution)
+
+    cols = list(range(bounds[0] // crop_size, bounds[2] // crop_size))
+    rows = list(range(bounds[1] // crop_size, bounds[3] // crop_size))
+
+    # Compute the land mask at the lattice of crop corners, in one vectorized pass.
+    corner_xs_px = np.array([col * crop_size for col in cols] + [bounds[2]])
+    corner_ys_px = np.array([row * crop_size for row in rows] + [bounds[3]])
```
APatrickJ: I think this might miss small islands that are in the land mask, given crops are about 20.5 km across.

We could refine the corner lattice by some factor to increase the likelihood that we'll catch these islands.

```python
LAND_SAMPLES_PER_CROP = 8
```

```suggestion
    step = crop_size // LAND_SAMPLES_PER_CROP
sample_xs_px = np.arange(bounds[0], bounds[2] + step, step)
    sample_ys_px = np.arange(bounds[1], bounds[3] + step, step)
```

Then later when we check if it's land:

```python
k = LAND_SAMPLES_PER_CROP
block = is_land[row_idx * k : row_idx * k + k + 1, col_idx * k : col_idx * k + k + 1]
if not block.any():
    continue
```

data/large_scale_embeddings/s2_s1.yaml
favyen2: OK I will make new rslearn release

rslp/large_scale_embeddings/predict_pipeline.py
favyen2: updated

rslp/large_scale_embeddings/predict_pipeline.py
favyen2: updated

rslp/large_scale_embeddings/write_jobs.py
favyen2: I think I fixed this in similar file in change finder, copied over the fix so it will intersect the WGS84 geometry with the current UTM zone bounds.

rslp/large_scale_embeddings/tiling.py
favyen2: updated

requirements.txt
favyen2: updated

rslp/large_scale_embeddings/tiling.py
favyen2: added LAND_STEP_SIZE to control how small of island you would like to capture (assuming the global land mask has it)

data/large_scale_embeddings/s2_s1.yaml
favyen2: updated

favyen2:
@claude can you review this PR?

APatrickJ (approved):
LGTM!

PR #336, opened 2026-07-23
Sentinel-1: raise marine infra filter to 200m and make it configurable

rpw1134:
The Sentinel-1 marine infra filter suppressed detections within 50m of a mapped structure, which is too tight — real infrastructure was still reaching the output as vessel detections. I cross-referenced two days of integration detections (33,105) against the same map the filter uses, and manually confirmed 6 infrastructure sites in the Skylight app. Across those, on-structure detections run out to ~200m (median 79m), with a clear gap before the next population starts. So: 0.05 → 0.2 km.

Also plumbed the value through like `score_threshold` already is (`SENTINEL1_INFRA_DISTANCE_KM` env var, per-request `infra_distance_km`) so the radius is tunable without rebuilding the image.

- Sentinel-2 (0.05) and Landsat (0.03) intentionally unchanged — analysis was S1 only.
- No perf change: the haversine scan is threshold-independent.
- `tests/unit` passes (25) plus 9 in `tests/unit/sentinel1_vessels`; ruff clean.


favyen2 (approved):

favyen2 (approved):
temporary fix sounds good to me, I'll look into the jsonargparse issue today

PR #344, opened 2026-08-28
NISAR vessel detection

favyen2:


rslp/nisar_vessels/scripts/create_predict_windows.py
```diff
@@ -0,0 +1,318 @@
...
+        geometry and collected_at), de-duplicated by item ID.
+    """
+    start = parse_time(start_time)
+    end = parse_time(end_time)
+    chunks = []
+    cur = start
+    while cur < end:
+        chunks.append((cur, min(cur + SEARCH_QUERY_CHUNK, end)))
+        cur += SEARCH_QUERY_CHUNK
+
```
APatrickJ, line 130: We might want to guard against an empty chunks list:

```python
if not chunks:
    logger.warning("empty chunk list")
    return []
```

rslp/nisar_vessels/scripts/create_dataset.py
```diff
@@ -0,0 +1,293 @@
...
+            task_status=task["status"],
+        ),
+    )
+    window.save()
+    window._data = dataset.window_data_storage_factory.create(window)
+
+    # Write the label layer with one point feature per annotation.
+    features = []
+    for annotation in annotations:
+        annotation_shp = shapely.from_wkt(annotation["geom_wkt"])
```
APatrickJ, line 224: Just in case, could skip annotations without a geometry (don't think this will happen in practice though):

```python
for annotation in annotations:
    if annotation["geom_wkt"] is None:
        logger.warning("skipping annotation %s with no geometry", annotation["id"])
        continue
    annotation_shp = shapely.from_wkt(annotation["geom_wkt"])
```

rslp/nisar_vessels/scripts/create_dataset.py
```diff
@@ -0,0 +1,293 @@
...
+        time_range=time_range,
+        options=dict(
+            split=split,
+            task_id=task["id"],
+            task_name=task["name"],
+            task_status=task["status"],
+        ),
+    )
+    window.save()
+    window._data = dataset.window_data_storage_factory.create(window)
```
APatrickJ: Could avoid the private `_data` access here by using the `Window` constructor:

```python
window = Window(
    storage=dataset.storage,
    ...
    data_factory=dataset.window_data_storage_factory,
)
window.save()
```

APatrickJ (approved):

rslp/nisar_vessels/scripts/create_dataset.py
```diff
@@ -0,0 +1,293 @@
...
+        int(dst_shp.bounds[3]) + 1,
+    )
+    time_range = (
+        parse_time(task["start_time"]) - TIME_BUFFER,
+        parse_time(task["end_time"]) + TIME_BUFFER,
+    )
+
+    # Include a portion of the task ID in the window name in case task names are not
+    # unique within the project.
+    window_name = f"{task['name']}_{task['id'][:8]}".replace(" ", "_")
```
APatrickJ: Might want to sanitize more than just spaces:

```python
safe_name = re.sub(r"[^A-Za-z0-9._-]", "_", task["name"])
window_name = f"{safe_name}_{task['id'][:8]}"
```

rslp/nisar_vessels/scripts/create_predict_windows.py
favyen2: I simplified it to just use `workers` instead of `min(workers, len(chunks))` for Pool(...) so it won't have error even if there are zero chunks.

rslp/nisar_vessels/scripts/create_dataset.py
favyen2: added this in

rslp/nisar_vessels/scripts/create_dataset.py
favyen2: added this in. I do see some parentheses in the current dataset so note that rerunning this script won't be idempotent wrt current dataset now (`NISAR_Group_B_2026-06-23_10N15W_(1)_0c3ccf74` -> `NISAR_Group_B_2026-06-23_10N15W__1__0c3ccf74`).

rslp/nisar_vessels/scripts/create_dataset.py
favyen2: fixed

PR #346, opened 2026-09-15
Add NISAR vessel prediction pipeline and FastAPI service

APatrickJ:
- Add a NISAR vessel detection pipeline and API service, based on Sentinel-1 but with no historical image and no attribute model
- Imagery arrives as HDF5 granules from the sat service, currently the only way to get NISAR imagery to the sidecar

The detector checkpoint is a TODO: no run has been published yet.


rslp/nisar_vessels/config.py
```diff
@@ -0,0 +1,37 @@
...
+# Host and port the FastAPI server binds to.
+NISAR_HOST = os.getenv("NISAR_HOST", "0.0.0.0")
+NISAR_PORT = int(os.getenv("NISAR_PORT", "5555"))
+
+# Default detector score threshold, overridable per request via the API.
+NISAR_SCORE_THRESHOLD = float(os.getenv("NISAR_SCORE_THRESHOLD", "0.7"))
+
+# Distance threshold for the near marine infrastructure filter in km
+# (0.2 km = 200 m if unset).
+INFRA_DISTANCE_THRESHOLD_KM = float(os.getenv("NISAR_INFRA_DISTANCE_KM", "0.2"))
```
favyen2: We use high threshold for Sentinel-1 because we use non terrain corrected scenes where the exact latitude/longitude corresponding to a pixel is not known, instead we just get something approximate. If NISAR data doesn't have this issue, this can be tightened to be similar to the threshold we use for Sentinel-2.

rslp/nisar_vessels/hdf5.py
```diff
@@ -0,0 +1,245 @@
...
+    x_dataset = group[X_COORDINATES_DATASET]
+    y_dataset = group[Y_COORDINATES_DATASET]
+    if x_dataset.shape[0] < 2 or y_dataset.shape[0] < 2:
+        raise ValueError(
+            f"Band {band} needs at least two coordinates per axis to derive a "
+            f"resolution, got {x_dataset.shape[0]}x{y_dataset.shape[0]}"
+        )
+
+    # Evenly spaced, so two coordinates suffice; the full arrays are hundreds of KB.
+    x_coordinates = x_dataset[:2]
```
favyen2: should there be some verification that the arrays are gridded here? as opposed to having varying spacing between the coordinates

rslp/nisar_vessels/hdf5.py
```diff
@@ -0,0 +1,245 @@
...
+    Returns:
+        the fill value, or None if the band does not declare one.
+    """
+    fill_value = group[band].attrs.get(FILL_VALUE_ATTRIBUTE)
+    if fill_value is None:
+        return None
+    return float(fill_value)
+
+
+def granule_to_geotiff(
```
favyen2, line 208: Is this granule H5 encoding specific to Skylight or is this how NASA provides the raw data? If it's Skylight-specific maybe this code should go in Skylight repository. This way the code here can be compatible with the raw data so users can easily execute it without needing to convert their data to H5.

rslp/nisar_vessels/predict_pipeline.py
```diff
@@ -0,0 +1,497 @@
...
+CROP_DECIBEL_RANGE = (-35.0, 5.0)
+
+# Matches the epsilon in rslearn's Sentinel1ToDecibels, so a zero or filled pixel lands
+# at -60 dB rather than negative infinity.
+DECIBEL_EPSILON = 1e-6
+
+# Crop size the detector tiles the scene into, overriding the config's training-time
+# value: a scene is far larger than a training window, and larger tiles mean fewer
+# forward passes. The overlap only has to exceed a vessel's footprint.
+PREDICT_CROP_SIZE = 512
```
favyen2: Might want to test these values on a few scenes and verify it is okay to increase the input crop size to this level.

rslp/nisar_vessels/hdf5.py
APatrickJ: ASF publishes NISAR products in HDF5, so this isn't specific to Skylight:

https://nisar-docs.asf.alaska.edu/data-format/

favyen2 (approved):

rslp/nisar_vessels/config.py
APatrickJ: Good point. The GCOV L2 product is [radiometrically terrain corrected](https://nisar-docs.asf.alaska.edu/products-overview/#level-2-geocoded-products), so I think we can tighten this to 0.05 to be similar to Sentinel-2.

rslp/nisar_vessels/hdf5.py
APatrickJ: Added a check that the spacings are all consistent

rslp/nisar_vessels/predict_pipeline.py
APatrickJ: I updated to 128/16 to match the training defaults

PR #347, opened 2026-09-16
Tile NISAR scenes into bounded windows to cap prediction memory

APatrickJ:
- Split each scene into overlapping tiles, one window each, so peak memory follows the tile rather than the granule
- Suppress detections duplicated across seams, reusing the suppression already applied within a window
- Read crops from the staged scene image, so a detection next to a seam still gets a full crop

Follows #323, but with overlapping rather than disjoint tiles so a vessel on a seam is whole in at least one.


favyen2 (approved):

rslp/utils/nms.py
```diff
@@ -15,6 +15,70 @@
...
+    distance_threshold of a higher-scoring one, so the best-scoring member of each
+    cluster survives. Returned indices are in ascending score order.
+
+    Args:
+        centers: (N, 2) array of (x, y) detection centers.
+        scores: (N,) array of detection scores.
+        distance_threshold: detections within this center distance are considered the
+            same object.
+        grid_size: cell size for the spatial index used to look up nearby centers.
+        indices: original indices for the rows of centers, defaulting to range(N).
```
favyen2: maybe add that this should be length N

rslp/nisar_vessels/predict_pipeline.py
```diff
@@ -209,88 +222,177 @@ def _get_scene_data(scene_id: str, grid: GranuleGrid) -> SceneData:
...
+    spans: list[tuple[int, int]] = []
+    start = low
+    while start < high:
+        end = min(start + tile_size, high)
+        spans.append((start, end))
+        if end >= high:
+            break
+        start += stride
+
+    if len(spans) > 1 and spans[-1][1] - spans[-1][0] < PREDICT_CROP_SIZE:
```
favyen2, line 285: I don't understand this part; if the overlap is 0 then won't this potentially leave part of the input image unprocessed?

favyen2: Oh nevermind I get it now, it's because unlike for the model inputs you have the flexibility to make windows that are slightly larger or smaller here so you're just making those windows bigger.

rslp/utils/nms.py
APatrickJ: Added and now consistent with `centers` and `scores`

rslp/nisar_vessels/predict_pipeline.py
APatrickJ: FWIW I added a test for zero overlap here

PR #349, opened 2026-09-21
Run Sentinel-1 and Landsat detection off the event loop

APatrickJ:
- Make the Sentinel-1 and Landsat detections endpoints sync, so a prediction runs in a worker thread instead of blocking the event loop
- Serialize inference behind a lock in both, matching sentinel2_vessels and nisar_vessels
- Note in each handler why it must not be async, and drop the test added in #348 that only checked for the keyword

A blocked event loop leaves the health probe unanswered until Kubernetes gives up and sends SIGTERM. Landsat had no mitigation at all; Sentinel-1 runs a second web worker to keep the probe alive, which this makes unnecessary.


tests/unit/landsat_vessels/test_api.py
```diff
@@ -0,0 +1,18 @@
+"""Unit tests for the Landsat vessel detection API request handling."""
+
+import inspect
+
+from rslp.landsat_vessels import api_main
+
+
+def test_detections_endpoint_is_not_async() -> None:
```
favyen2: I think a test that verifies it isn't blocked makes sense but this test just basically checks that it doesn't have "async" keyword which doesn't seem necessary. So I would suggest to remove or to make it actually verify the event loop not getting blocked.

favyen2 (commented):
The code changes look good but I think it's better to make a note about not using async in the code instead of having a test that effectively verifies it is not async.

tests/unit/landsat_vessels/test_api.py
APatrickJ: Yeah I agree about this. I just dropped the test and added comments.

favyen2 (approved):

